PR #17, opened 2021-12-01
Allow logging to multiple output

marcosnav:
Connected to #19 

- Allow logging to multiple output
- Remove the responsibility to create default log files from this lib (products should handle that)

### DIFF
--- a/pkg/logger/output_builder.go
+++ b/pkg/logger/output_builder.go
@@ -73,40 +73,43 @@ func (o osFileSystem) OpenFile(name string, flag int, perm os.FileMode) (*os.Fil
 }
 
 type OutputBuilder interface {
-	Build(output LogOutputType, logFilePath, defaultLogFilePath string) io.Writer
+	Build(output LogOutputType, logFilePath string) (io.Writer, error)
 }
 
 type outputBuilder struct {
 	logCategory LogCategory
 	fileSystem  FileSystem
+	defaultFile string
 }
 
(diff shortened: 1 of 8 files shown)


pkg/logger/output_builder.go
```diff
@@ -73,40 +73,43 @@ func (o osFileSystem) OpenFile(name string, flag int, perm os.FileMode) (*os.Fil
...
 
 type OutputBuilder interface {
-	Build(output LogOutputType, logFilePath, defaultLogFilePath string) io.Writer
+	Build(output LogOutputType, logFilePath string) (io.Writer, error)
 }
 
 type outputBuilder struct {
 	logCategory LogCategory
 	fileSystem  FileSystem
+	defaultFile string
```
jonyoder, line 82: Should this be named `defaultFile` or simply `logFile`?

jonyoder (commented):
I like this a lot. One question about variable naming. Also, when this merges, we'll want to bump the version and update the Connect PR with the logic to create the default log directories if necessary.

pkg/logger/output_builder.go
marcosnav: I think `defaultFile`, since the purpose of this one is to have the default destination file in hand in case a provided custom file errors, knowing that the default logging directories and files should have been created at product installation.

jonyoder (approved):

jonyoder:
@marcosnav feel free to merge this when you're ready. I think the RSPM team is anxious for this feature!

PR #20, opened 2021-12-20
Debug logging extras

marcosnav:
- Use default logger within output builder (to use the latest configured logger whenever possible)
- Fixed DefaultLoggerMock (missing method), and return interfaces on methods for flexibility

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -24,7 +24,7 @@ type Logger interface {
 	SetLevel(level LogLevel)
 	SetReportCaller(bool)
 
-	// TODO: remove this interface when the Connect migration process to the new logging standard is complete.
+	// TODO: remove this interface when the migration process to the new logging standard is complete.
 	// It is being added just to help migrate packages that inject the Logger interface inside another packages.
 	DeprecatedLogger
 }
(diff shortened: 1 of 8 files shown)


eduardocfalcao (approved):

pkg/logger/logger.go
```diff
@@ -23,6 +23,10 @@ type Logger interface {
 	WithFields(fields Fields) Logger
 	SetLevel(level LogLevel)
 	SetReportCaller(bool)
+
+	// TODO: remove this interface when the Connect migration process to the new logging standard is complete.
```
jonyoder, line 27: Consider removing the Connect reference while keeping the comment.

pkg/logger/logger_impl.go
```diff
@@ -186,6 +191,11 @@ func (l logrusEntryWrapper) SetReportCaller(flag bool) {
 	l.Logger.SetReportCaller(flag)
 }
 
+// TODO: remove this function when the Connect migration process to the new logging standard is complete.
```
jonyoder, line 194: Here, and above on line 147, consider removing the Connect reference while keeping the comment.

jonyoder (commented):
Looks good! I saw three comments that referenced Connect that we might want to make more generic.

PR #21, opened 2021-12-20
Adds Panicf function to Logger interface

eduardocfalcao:
Adding `Panicf` function to the Logger interface.
This function was added in a commit when the logger package was inside Connect.


marcosnav (approved):
LGTM

jonyoder (approved):
🍰

PR #23, opened 2021-12-23
Renames logger package to rslog

eduardocfalcao:
This PR attempts to rename the `logger` package to `rslog`.

We thought it can good for the systems that use this library and have an internal package called logger. 
It happened with us on Connect, and renaming this package makes sense for us.


aronatkins (approved):
At some point, we'll want to think about versioning the library; this type of change would require a major version change.

marcosnav (approved):
LGTM :)

eduardocfalcao:
> At some point, we'll want to think about versioning the library; this type of change would require a major version change.

@aronatkins 
Yup! you totally right. But at the moment, we are good following with this. But we should address this soon.

jonyoder:
@aronatkins @eduardocfalcao we already have basic versioning in place. Currently we're simply pushing a tag to bump the version, and then you can constrain the version in your `go.mod` as necessary. See the releases page for the current prereleases.

jonyoder (approved):

Issue #22, opened 2021-12-22
Explore Debug Logging without DI


jonyoder:
We should explore options for using debug loggers without requiring dependency injection.

- [x] Provide easier way for Connect/RSPM to use debug logging in their code that both teams will like

jonyoder:
Update: Currently, we're leaning towards eliminating the global state in this library. Each application that uses the library can decide how to instantiate loggers and whether to store them as globals or use DI.

PR #24, opened 2022-01-06
Use a single, non-replaceable default logger

jonyoder:
Connected to #22. This PR:

* Removes the `SetDefaultLogger` method. Instead of replacing the default logger with a new reference, simply use the `Set*` methods to update the existing default logger as needed.
* Adds the methods `SetFormatter` and `Tracef` to the logger interface.
* Adds the method `Tracef` to the debug logger interface.
* Adds a public global `DefaultLoggerFactory` that can be set in an `init()` method to provide the `rslog` package with a different default logger. If this is unset, `rslog` will automatically create its own factory and default logger. See `root.go` in the example command directory for an example.
* Removes the logger interface's `Copy` method. We no longer support copying a logger since it makes SIGHUP support impossible.
* Removes the logger interface's `SetReportCaller` method. Since logrus reports the first non-logrus line of code, this feature was useless since it always simply reported a line of code in the `rslog` package.
* Updates the debug logger to use the underlying `defaultLogger` rather than copying an existing logger.
* Updates the example command with a `debug` command and a new global `--level` flag. 
* Adds a `Compare(l LogLevel)` method to the `LogLevel` struct so you can compare log levels based on the underlying logrus values.

Using the example commands:

```sh
just build
./out/testapp debug --message="hi jonathan" --level=TRACE
./out/testapp debug --message="hi jonathan" --level=DEBUG
./out/testapp log --message="hi jonathan" --level=INFO
./out/testapp log --message="hi jonathan" --level=WARN
```

### DIFF
--- a/pkg/rslog/debug/debug.go
+++ b/pkg/rslog/debug/debug.go
@@ -104,32 +104,30 @@ func Enabled(region ProductRegion) bool {
 type DebugLogger interface {
 	Enabled() bool
 	Debugf(msg string, args ...interface{})
+	Tracef(msg string, args ...interface{})
 	WithFields(fields rslog.Fields) DebugLogger
 	WithSubRegion(subregion string) DebugLogger
 }
 
 type debugLogger struct {
 	rslog.Logger
-	region ProductRegion
+	region  ProductRegion
+	enabled bool
 }
 
 // NewDebugLogger returns a new logger which includes
 // the name of the debug region at every message.
-func NewDebugLogger(region ProductRegion, lgrIn rslog.Logger) *debugLogger {
-	lgr := lgrIn.Copy()
-
-	if Enabled(region) {
-		lgr.SetLevel(rslog.DebugLevel)
-		lgr.SetReportCaller(true)
-	}
+func NewDebugLogger(region ProductRegion) *debugLogger {
+	lgr := rslog.DefaultLogger()
 
 	entry := lgr.WithFields(rslog.Fields{
 		"region": regionNames[region],
 	})
 
 	dbglgr := &debugLogger{
-		Logger: entry,
-		region: region,
+		Logger:  entry,
+		region:  region,
+		enabled: Enabled(region),
 	}
 
 	registerLoggerCb(region, dbglgr.enable)
@@ -144,6 +142,18 @@ func registerLoggerCb(region ProductRegion, cb func(bool)) {
 // Enabled returns true if debug logging is enabled for this rslog.
 func (l *debugLogger) Enabled() bool {
 	return Enabled(l.region)
+}
+
+func (l *debugLogger) Debugf(message string, args ...interface{}) {
+	if l.enabled {
+		l.Logger.Debugf(message, args...)
+	}
+}
+
+func (l *debugLogger) Tracef(message string, args ...interface{}) {
+	if l.enabled {
+		l.Logger.Tracef(message, args...)
+	}
 }
 
 // Set fields to be logged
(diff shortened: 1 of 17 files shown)


pkg/rslog/debug/debug.go
```diff
@@ -104,32 +104,30 @@ func Enabled(region ProductRegion) bool {
...
 // the name of the debug region at every message.
-func NewDebugLogger(region ProductRegion, lgrIn rslog.Logger) *debugLogger {
-	lgr := lgrIn.Copy()
-
-	if Enabled(region) {
-		lgr.SetLevel(rslog.DebugLevel)
-		lgr.SetReportCaller(true)
-	}
+func NewDebugLogger(region ProductRegion) *debugLogger {
+	lgr := rslog.DefaultLogger()
```
marcosnav, line 121: It is looking good to me. One issue we currently have, noticed by @eduardocfalcao , is that sometimes `NewDebugLogger()` is called before the default logger is properly initialized(due to order of package imports in Connect), leaving a logger instance that is not in sync with the configuration set.

We are thinking this could be solved if `rslog.DefaultLogger()` returns a pointer to the default logger, instead of the value. WDYT?

marcosnav:
We'll want to bump a new version of the package when this changes land. But before that, we also want to rename `rslog/loggertest` to `rslog/rslogtest`. It would be good to push both things in the same tag. 🙂

pkg/rslog/debug/debug.go
jonyoder: Yes, that's one of the primary purposes of this PR. 😁 I still need to do some further testing, but I think the singleton default logger (that cannot be replaced) will solve that problem.

marcosnav: Oh, yes, returning the default logger instead of a copy should solve that problem, I got confused 😅

jonyoder:
I will make the change to `rslogtest` in this PR while I'm here.

pkg/rslog/logger_impl.go
```diff
@@ -187,53 +167,124 @@ func (l logrusEntryWrapper) WithFields(fields Fields) Logger {
 	return logrusEntryWrapper{Entry: e}
 }
 
-func (l logrusEntryWrapper) SetReportCaller(flag bool) {
-	l.Logger.SetReportCaller(flag)
-}
```
marcosnav, line 192: I think this change could impact our debug logs, we use `SetReportCaller` for debug logs to be able to come with the context of the caller (source code file and line num). We might want to keep this, but being a single instance in use, this will impact other methods as well like `Infof/Warnf` and not just `Debugf/Tracef`

aronatkins: I am OK if we abandon source file/lineno information from debug logs.

pkg/rslog/logger.go
```diff
@@ -97,6 +102,18 @@ func (o *LogLevel) UnmarshalText(text []byte) (err error) {
 	return fmt.Errorf("invalid Log Level value '%s'. Allowed values are %v", text, values)
 }
 
+func (o *LogLevel) Compare(level LogLevel) int {
```
jonyoder, line 105: @marcosnav @eduardocfalcao I just added this method this morning so we can check for a compatible log level in Connect when debug regions are enabled. I also added a corresponding test in `logger_test.go`.

jonyoder:
Also added one more commit that includes an example service you can run to prove that logger reconfiguration works: https://github.com/rstudio/platform-lib/pull/24/commits/72b5b6a434fc88c3712060a6a8cca4b05b6809a4.

pkg/rslog/logger_impl.go
scottmmjackson: Abandoning file/lineno information is probably more important given that we're headed in the direction of making it customer facing. Gives up information about the underlying source that we might not want to.

jonyoder: @scottmmjackson And now I see I totally misread your comment. 😆 You're right, that can be a small risk!

jonyoder:
@marcosnav @eduardocfalcao Are you ok with merging this one now?

marcosnav (approved):

marcosnav:
> @marcosnav @eduardocfalcao Are you ok with merging this one now?

@jonyoder Yes, I'm okay with it :)

Issue #25, opened 2022-01-10
Add NOTICE.md


jonyoder:
We should add a NOTICE file that enumerates third-party depenencies.

PR #26, opened 2022-01-10
Licenses/Versioning

jonyoder:
Connected to #25.

* Adds a `just licenses` target that updates `NOTICE.md` with a list of third-party Go licenses.
* Updates README with notes on licenses and versioning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,13 +75,20 @@ use the `just licenses` target. This requires Python 3.
 
 ## Versioning
 
-Follow semantic versioning guidelines. To release a new version, we simply
+Follow [semantic versioning](https://go.dev/doc/modules/version-numbers) guidelines. To release a new version, we simply
 create and push a tag.
 
 ```shell
 git tag v0.1.2
 git push origin v0.1.2
 ```
+
+## Release Checklist
+
+- [ ] Ensure build passes on `main` branch.
+- [ ] Update `README.md` as necessary.
+- [ ] Update `NOTICE.md`. See [Licenses](#licenses).
+- [ ] Tag the new version. See [Versioning](#versioning)
 
 ## Badges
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -16,12 +16,6 @@ You can optionally add a bit of extra structure to your internal packages to sep
 
 Library code that's ok to use by external applications (e.g., `/pkg/mypubliclib`). Other projects will import these libraries expecting them to work, so think twice before you put something here.
 
-### `/vendor`
-
```
kfeinauer, line 20: This isn't true anymore?

kfeinauer: Oh, I guess this is just taken care of by the make commands now?

jonyoder: Vendoring third-party libs is apparently an antipattern for shared modules. I think it's only recommended for application codebases.

kfeinauer (approved):
Very cool license finder script! We might want to eventually port this to Launcher since we are pulling in Go packages now.

README.md
kfeinauer: Could you link some documentation on that? Curious why that would be since this repo should stand/build on its own.

jonyoder: I've actually had trouble tracking down documentation on this, but the go layout template I used for this project actually states it here https://github.com/golang-standards/project-layout/tree/master/vendor:

> Don't commit your application dependencies if you are building a library.

I'm not clear on what happens with the `vendor` directory in a third-party lib, but I'm guessing it's not used by a consuming application. So, if we're vendoring our code for the purpose of testing, we could have passing builds while some consumer's build would fail due to some other third-party issue. In addition, the consuming application's dependencies may end up with some conflicts with ours.

aronatkins:
Reference to original: https://github.com/rstudio/connect/blob/7d7c44c04e66ffa8e3f135f3319cdc06a498d32c/scripts/go-licenses.py

jonyoder:
Yes, many thanks to Connect for the license finder script!

jonyoder:
Merged with #33.

PR #27, opened 2022-01-11
Port notify package

jonyoder:
Ports the notify package from a product. This code was ported verbatim with the one exception that the `notify` package was renamed to `rsnotify`.


kfeinauer (approved):

jonyoder:
Merged with #33.

PR #30, opened 2022-01-11
Add lib/pq listener

jonyoder:
Adds a beta `lib/pq` listener implementation. So far, this has only been unit tested; we need further testing to prove that it works reliably.

### DIFF
--- a/pkg/rsnotify/pqlistener/pqlistener.go
+++ b/pkg/rsnotify/pqlistener/pqlistener.go
@@ -1,0 +1,204 @@
+package pqlistener
+
+/* pqlistener.go
+ *
+ * Copyright (C) 2021 by RStudio, PBC
+ * All Rights Reserved.
+ *
+ * NOTICE: All information contained herein is, and remains the property of
+ * RStudio, PBC and its suppliers, if any. The intellectual and technical
+ * concepts contained herein are proprietary to RStudio, PBC and its suppliers
+ * and may be covered by U.S. and Foreign Patents, patents in process, and
+ * are protected by trade secret or copyright law. Dissemination of this
+ * information or reproduction of this material is strictly forbidden unless
+ * prior written permission is obtained.
+ */
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"reflect"
+	"time"
+
+	"github.com/lib/pq"
+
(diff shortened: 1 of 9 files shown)


pkg/rsnotify/pqlistener/pqlistener.go
```diff
@@ -0,0 +1,204 @@
...
+	l.ip = ""
+	l.conn, err = l.factory.NewListener()
+	if err != nil {
+		return fmt.Errorf("error acquiring native connection: %s", err)
+	}
+
+	// Get the connection's IP
+	l.ip, err = l.factory.IP()
+	if err != nil {
+		return
```
kfeinauer, line 140:
```suggestion
		return fmt.Errorf("error determining listener IP address: %s", err)
```
Do we want to use this idiom here?

pkg/rsnotify/pqlistener/pqlistener.go
```diff
@@ -0,0 +1,204 @@
...
+		errs <- fmt.Errorf("error unmarshalling JSON: %s", err)
+		return
+	}
+	if unmarshaller, ok := l.unmarshallers[input.Type()]; ok {
+		err = unmarshaller(input, payloadBytes)
+		if err != nil {
+			errs <- fmt.Errorf("error unmarshalling with custom unmarshaller: %s", err)
+			return
+		}
+		l.Debugf("Unmarshalled notification of type %d with registered unmarshaller", input.Type())
```
kfeinauer, line 184: Should we mention the unmarshaller type here? Might be valuable information.

kfeinauer (approved):

pkg/rsnotify/pqlistener/pqlistener.go
jonyoder: Will do!

pkg/rsnotify/pqlistener/pqlistener.go
jonyoder: An unmarshaller is a function, so I'm not sure that would be useful: https://github.com/rstudio/platform-lib/blob/4cf2da70d4cbc461b6bd0480e347ae78a718cd89/pkg/rsnotify/listener/listener.go#L32.

That said, maybe we can eventually work on a way to name the unmarshallers for better reporting.

jonyoder:
Merged with #33.

PR #34, opened 2022-01-13
Create a CapturingLogger following the logging standards

eduardocfalcao:
This PR attempts to create a new `CapturingLogger` that follows the logging standards.

So it's being added:
 - Deprecating all the old capturing, recording, or kind of logger.
 - Creating a new CapturingLogger, that follows the new standard.
 - Creating a CompositeLogger, making it possible to create a group of loggers to substitute loggers like `CapturingPassthroughLogger`.
 - Removing the `CapturingPassthroughLogger`. Since it was being in use just in one place, that will be replaced by the new `CapturingLogger`.
 - Adding unit tests for the new loggers
 - Adding new examples of usage at the sample app.


jonyoder (changes requested):
Looks great! I noticed a few small things to fix, but nothing big.

marcosnav (approved):

PR #35, opened 2022-01-13
Split into multiple Go modules

jonyoder:
To avoid dependency issues, this splits the packages into multiple Go modules. For a similar layout, see the AWS SDK https://github.com/aws/aws-sdk-go-v2.


eduardocfalcao (approved):
Great work! :+1:

PR #36, opened 2022-01-14
Make each rsnotify listener implementation a module

jonyoder:
* Makes each `Listener` implementation a module of its own. This will make it easy to import only the listener types you need.
* Moves all of the listener factory implementations into the correct `listeners` implementation packages.
* Refactors some naming per Go recommendations. For example "local.LocalListener" becomes "local.Listener".


pkg/rsnotify/listenerfactory/listenerfactory.go
```diff
@@ -0,0 +1,27 @@
+package listenerfactory
```
jonyoder, line 1: The code here was moved from `factory/listenerfactory.go`. None of it has changed, apart from removing the implementations and moving them to the correct listener implementation packages.

pkg/rsnotify/listeners/local/factory.go
```diff
@@ -0,0 +1,28 @@
+package local
```
jonyoder, line 1: Moved from the old factory. This code remains basically unchanged.

pkg/rsnotify/listeners/local/locallistener.go
```diff
@@ -13,34 +13,34 @@ import (
...
+	guid     string
+	provider *ListenerProvider
 
 	// If this channel is non-nil, listening is deferred until this channel
 	// is signaled or closed. Currently used in test only.
 	deferredStart chan struct{}
 }
 
-type LocalListenerFactory struct {
+type ListenerProvider struct {
```
jonyoder, line 28: This name conflicted with the factory and `Provider` seemed appropriate.

pkg/rsnotify/listeners/postgrespgx/factory.go
```diff
@@ -0,0 +1,43 @@
+package postgrespgx
```
jonyoder, line 1: Moved from the old factory. This code remains basically unchanged.

pkg/rsnotify/listeners/postgrespq/factory.go
```diff
@@ -0,0 +1,41 @@
+package postgrespq
```
jonyoder, line 1: Moved from the old factory. This code remains basically unchanged.

tylfin (approved):
This all makes sense for me.

For projects that require both, could be dependency requirements that demand they be upgraded in step anyways?

PR #42, opened 2022-01-19
Change CompositeLogger to return a proper value from the Writer function

eduardocfalcao:
I am adding the `CompositeLogger` capability to return multiple writers on the `Writer` function on this PR. 

To make it possible:
- I created a new struct, `multiWriteCloser` (to handle a `WriteCloser` slice as just one `WriteCloser`)
- Changed the return from the `Logger.Writer` function to `io.WriteCloser` (where it will be used, only need this interface).

### DIFF
--- a/examples/cmd/testapp/cmd/log.go
+++ b/examples/cmd/testapp/cmd/log.go
@@ -6,6 +6,7 @@ import (
 	"log"
 
 	"github.com/rstudio/platform-lib/pkg/rslog"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
@@ -20,7 +21,7 @@ func init() {
 	LogCmd.Flags().StringVar(&message, "message", "default message", "The message to log.")
 	CaptureLogCmd.Flags().BoolVar(&withMetadata, "withMetadata", false, "The option to turn on or off the metadata in the capturing logger")
 
-	RootCmd.AddCommand(LogCmd, CaptureLogCmd, CompositeLogCmd)
+	RootCmd.AddCommand(LogCmd, CaptureLogCmd, CompositeLogCmd, TestCompositeWriter)
 }
 
 var LogCmd = &cobra.Command{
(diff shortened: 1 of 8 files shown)


examples/cmd/testapp/cmd/log.go
```diff
@@ -61,9 +62,46 @@ var CompositeLogCmd = &cobra.Command{
...
+		logr.Out = w
+		finalLogger := rslog.LoggerImpl{
+			Logger: logr,
+		}
+
+		finalLogger.Infof("Testing multiwriter message 1")
+		finalLogger.Infof("Testing multiwriter message 2")
+		finalLogger.Infof("Testing multiwriter message 3")
+
+		w.Close()
```
eduardocfalcao, line 104: I'm having intermittent behavior here. Sometimes, the last message isn't logged in one or both outputs defined.

If I remove this close function call, it works smoothly. But as it can be called, I decided to point here to get some advice that could avoid that. On `Connect,` it will be called from inside the `util.CloseScope`. So I don't know if it is a problem that we can skip for a while.

jonyoder: This is because the writer returned by logrus scans for input in a goroutine: https://github.com/sirupsen/logrus/blob/master/writer.go#L51

I haven't looked at this in detail, but it looks like it'll be a bit difficult to be deterministic about when the writes actually happen.

jonyoder: To demonstrate some of the oddness, you can modify the multiwriter close with some additional logs and then run the example command:

```go
func (mw *multiWriteCloser) Write(p []byte) (n int, err error) {
	fmt.Printf("I have %d writers\n", len(mw.writeClosers))
	for i, w := range mw.writeClosers {
		fmt.Printf("handling writer %d\n", i)
		n, err = w.Write(p)
		if err != nil {
			return
		}
		if n != len(p) {
			err = io.ErrShortWrite
			return
		}
	}
	return len(p), nil
}
```

eduardocfalcao: Anyway, after new tests, I saw that I got the same behavior even after removing the `w.Close()` function call. 
Also, Logrus already handles when to close the writer. So maybe we don't need to care about it. 

At least on connect, it might not be a problem because the main process (who keeps the writer) will keep running after all stuff that needs the writer.

jonyoder: What's your use case in Connect? If you need to be deterministic about writing, then I think you should avoid using the `Writer()` method from logrus. But it sounds like it's likely not a big deal in this case.

aronatkins: @eduardocfalcao and I just chatted. We have this "stream to logger" flow that is used for our interactive process output. Really, it is only something that's in play for certain very narrow configurations. The use of `Writer()` from logrus is in-code, but not actively enabled.

We could definitely rewrite this integration to use a stream parser that logs each received line, rather than rely on the logger to do this parsing; we use this approach elsewhere. If we use a separate adapter, we could take control of its lifecycle and avoid the `Close()` race condition entirely.

The `Close()` race is unlikely to matter a ton in a real application, but is definitely confusing when seen in a demo application like this one.

eduardocfalcao: After conversing with Marcos, we decided not to follow up with the Writer function and handle it differently where we need it on Connect. I will close this PR, and I will remove the `Writer` function on Marcos PR. #39

PR #44, opened 2022-01-21
Creating a discarding logger following the new logging standards

eduardocfalcao:
Because some parts of Connect ignore logging, we still need a logger that doesn't log the messages but should implement the logger interface. 

So, this pr aims to:
- Create an easy discarding logger, that implements all functions from the `Logger` interface.
- It also renames the function `NewCompositeLogger` to `ComposeLoggers`.

### DIFF
--- a/pkg/rslog/composite_logger.go
+++ b/pkg/rslog/composite_logger.go
@@ -10,7 +10,7 @@ type CompositeLogger struct {
 
 var _ Logger = &CompositeLogger{}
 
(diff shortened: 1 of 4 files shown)


jonyoder (approved):

jonyoder:
FYI, there are vet errors: https://github.com/rstudio/platform-lib/runs/4900755464?check_suite_focus=true

jonyoder (changes requested):
Fix vet errors before merging.

marcosnav (approved):

PR #47, opened 2022-01-26
Go module for rsstorage

jonyoder:
Builds on #46 by changing the package name to `rsstorage`, adding a `go.mod`, correcting a few comments, and adding a README. This build will also fail, which is expected.


jonyoder:
Merged with #52.

PR #48, opened 2022-01-26
Add MinIO to docker-compose for S3 integration tests

jonyoder:
Builds on #46 and #47 by adding MinIO to the docker-compose configuration. This gets the unit tests all passing.

PR #49, opened 2022-01-26
Split storage servers into separate modules

jonyoder:
Builds on #46, #47, and #48 by splitting the storage server implementations into separate go modules.

* Adds `local`, `postgres`, and `s3storage` go modules for the three server implementations.
* Adds an `integration_test` go module to keep the heavy unit tests packaged separately. This will help applications that don't need to import all of the storage implementations.
* Makes some factory code private since we don't really want apps using code that imports all of the server implementations.
* Adds README files for each implementation.


pkg/rsstorage/chunks.go
```diff
@@ -267,3 +268,31 @@ func (w *defaultChunkUtils) tryChunkRead(attempts int, chunkIndex uint64, addres
 
 	return true, nil
 }
+
+func FilterChunks(input []PersistentStorageItem) []PersistentStorageItem {
```
jonyoder, line 272: This function was moved from another file. It makes more sense to have it here.

pkg/rsstorage/integration_test/integration.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/jackc/pgx/v4/pgxpool"
+
+	"github.com/rstudio/platform-lib/pkg/rsstorage"
+	"github.com/rstudio/platform-lib/pkg/rsstorage/servers/file"
+	"github.com/rstudio/platform-lib/pkg/rsstorage/servers/postgres"
+	"github.com/rstudio/platform-lib/pkg/rsstorage/servers/s3server"
+)
+
+// Simply wraps getStorageServerAttempt, fatally erring if something goes wrong
+func GetStorageServer(cfg *rsstorage.Config, class string, destination string, waiter rsstorage.ChunkWaiter, notifier rsstorage.ChunkNotifier, pool *pgxpool.Pool, cstore rsstorage.PersistentStorageStore, debugLogger rsstorage.DebugLogger) (rsstorage.PersistentStorageServer, error) {
```
jonyoder, line 17: These two functions were previously public factory methods that helped to set up storage in RSPM. However, they're not ideal since they reference all the storage server implementations and hence require importing all implementations into your application. Moving them to the `integration_test` package since they're still useful for tests.

jonyoder:
Merged with #52.

PR #53, opened 2022-01-28
Zack debug module

zackverham:
### Intent

- Fixes a package naming collision with connect by moving the `debug` code up to the core `rslog` package. 
- Updates `.gitignore` to include goland files
- The shell scripts utilized by `just` were using a non-standard `xargs` extension - updated to (hopefully) be flexible across multiple environments.

### Approach

Core change is simply moving code between packages. The primary complexity came up in the unit tests, which rely on a single-operation mutex to set the default logger. A single package-wide mocked logger is now used for `rslogtest` which also required some additional mocking. 

### Automated Tests

Still uses same test suite - core logic should remain unchanged. 

### QA Notes

- We will want to validate that the just commands which exercise the shell scripts work in multiple environments. 

- In the review, I'd appreciate emphasis on debug_test.TestNewDebugLogger, where I had to introduce some specific mocking logic.  

### Checklist

I don't believe there are particularly complex steps aside from code review. Actually integrating this change into Connect will require coordination.


examples/cmd/testnotify/cmd/notify.go
```diff
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/rstudio/platform-lib/pkg/rslog"
```
jonyoder, line 9: This import should be grouped correctly.

examples/cmd/testnotify/cmd/root.go
```diff
@@ -7,24 +7,23 @@ import (
 	"time"
 
 	"github.com/lib/pq"
+	"github.com/rstudio/platform-lib/pkg/rslog"
```
jonyoder, line 10: import grouping.

pkg/rslog/logger_impl.go
```diff
@@ -174,7 +174,6 @@ func (l logrusEntryWrapper) Logf(msg string, args ...interface{}) {
 
 var defaultLogger Logger
 var once = &sync.Once{}
-var mutex sync.RWMutex
```
jonyoder, line 177: Would it make more sense to rename the mutex in `debug.go` to `debugMutex` so we can still keep them separate?

jonyoder (changes requested):
A couple of import formatting nits. Other than that, I wonder if we should rename one of the mutexes so they stay separate.

pkg/rslog/logger_impl.go
zackverham: ah, yes - that does make sense to me. Good call.

jonyoder (approved):

marcosnav (approved):
✨

PR #56, opened 2022-02-01
LoggerMock.Fatal() bug fixes

zackverham:
# Intent 

Fixes two bugs with the `Fatal` function in the mocked logger: 

1. `Fatal` wasn't adding its logging operation to the mock's internal call stack, causing `LastCall()` assertions to fail 
2. Because `Fatal` has a different method signature than the other formatted logging statements, `AllowAny` was failing to register the correct method signature. 

# Automated Tests

I added a new test to validate the change and confirm that the rest of the logging functions use the call stack as expected. The test also implicitly confirms `AllowAny` behavior by allowing all logging operations. 

# Miscellaneous

I added a todo comment in `LoggerMock.AllowAny` which I think could be worth some discussion - the design could totally be the right approach, but I'm just curious why we use string parameters rather than an enumerated type here.

### DIFF
--- a/pkg/rslog/rslogtest/mocks.go
+++ b/pkg/rslog/rslogtest/mocks.go
@@ -40,7 +40,12 @@ type LoggerMock struct {
 // allowing to be called with any message and arguments
 func (m *LoggerMock) AllowAny(methods ...string) {
 	for _, method := range methods {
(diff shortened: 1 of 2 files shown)


marcosnav (approved):
I agree that it might be good to try out having types/enums for the allowed methods (`AllowAny`). Not required in this PR 🙂

PR #58, opened 2022-02-16
Add Registered Type Matchers for Listeners

jonyoder:
Historically, when you create a listener to receive notifications on a channel, that listener is bound to a single type. This makes it difficult to reuse a single DB connection for different notification types.

This PR adds the ability to register multiple types for a single listener. The type matcher provides the listener with the ability to listen for multiple types reliably.

### DIFF
--- a/pkg/rsnotify/listener/listener.go
+++ b/pkg/rsnotify/listener/listener.go
@@ -2,8 +2,10 @@ package listener
 
 // Copyright (C) 2022 by RStudio, PBC.
 
+import "encoding/json"
+
 const (
-	// Sizing for the channel to prevent blocking when distributing
+	// MaxChannelSize sizes for the channel to prevent blocking when distributing
 	// notifications to listeners.
 	MaxChannelSize = 100
 )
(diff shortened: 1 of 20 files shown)


pkg/rsnotify/listeners/postgrespgx/pgxlistener.go
```diff
@@ -170,18 +170,40 @@ func (l *PgxListener) Debugf(msg string, args ...interface{}) {
...
+	// Convert postgres message payload to byte array
 	payloadBytes := []byte(n.Payload)
-	err := json.Unmarshal(payloadBytes, input)
+
+	// Unmarshal the payload to a raw message
+	var tmp map[string]*json.RawMessage
+	err := json.Unmarshal(payloadBytes, &tmp)
+	if err != nil {
+		errs <- fmt.Errorf("error unmarshalling raw message: %s", err)
+	}
```
nodivbyzero, line 185: Don't you want to return in case of error?

pkg/rsnotify/listener/listener.go
```diff
@@ -44,7 +26,7 @@ type Listener interface {
 }
 
 type Logger interface {
-	Debugf(msg string, args ...interface{})
+	Debugf(msg string, argslistenerfactory ...interface{})
```
jonyoder, line 29: This was a copy/paste mistake. I'll fix it.

tylfin (approved):

PR #61, opened 2022-02-28
Prevent hanging behaivour when calculating file storage disk usage

tylfin:
This issue is intended to prevent infinite hangs due to filesystem locking issues. Here:

- [x] Add `walkTimeout` to `diskUsage` and `enumerate` functions
- [x] Add new internal `enumerate` function to make testing the walk easier
- [x] Run the `filesystem.Walk` in a separate goroutine using channels to perform two-way communication
- [x] Add new `walkTimeout` argument to `NewFileStorageServer`

~If this looks good, I'll also update the `Enumerate()` function similarly.~

**Note:** The point of the `cacheTimeout` was to prevent the `diskUsage` function from hanging longer than our scheduler if the file directory was massive. The new `walkTimeout` variable is meant to be a sanity check, e.g. calculating the size of a single file should never take more than say 15 - 30s.

This will be the first part of: https://github.com/rstudio/package-manager/issues/7475, I'll also make a change in RSPM to stop blocking start-up until after the file storage calculation finishes.

### DIFF
--- a/pkg/rsstorage/servers/file/server.go
+++ b/pkg/rsstorage/servers/file/server.go
@@ -144,7 +144,8 @@ func diskUsage(duPath string, cacheTimeout, walkTimeout time.Duration) (size dat
 	defer close(stop)
 
 	sizeChan := make(chan datasize.ByteSize)
-	errChan := make(chan error)
+	// errChan should have a buffer of two items to prevent deadlock between `<-stop` and `errChan<-err`
+	errChan := make(chan error, 2)
 
 	go func(stop <-chan bool, sizeChan chan<- datasize.ByteSize, errChan chan<- error) {
 		err = filepath.Walk(duPath, func(path string, info os.FileInfo, err error) error {
@@ -348,7 +349,8 @@ func enumerate(dir string, walkTimeout time.Duration) ([]types.StoredItem, error
 	defer close(stop)
 
 	itemChan := make(chan *types.StoredItem)
-	errChan := make(chan error)
+	// errChan should have a buffer of two items to prevent deadlock between `<-stop` and `errChan<-err`
+	errChan := make(chan error, 2)
 
 	go func(stop <-chan bool, itemChan chan<- *types.StoredItem, errChan chan<- error) {
 		err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
--- a/pkg/rsstorage/servers/file/server.go
+++ b/pkg/rsstorage/servers/file/server.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"io/ioutil"
 	"log"
 	"os"
@@ -360,7 +361,7 @@ func enumerate(dir string, walkTimeout time.Duration) ([]types.StoredItem, error
 		defer close(itemChan)
 		defer close(errChan)
 
-		err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		err := filepath.WalkDir(dir, func(path string, info fs.DirEntry, err error) error {
 			if err != nil {
 				log.Printf("Error enumerating storage for directory %s: %s", dir, err)
 				return nil
--- a/pkg/rsstorage/servers/file/server.go
+++ b/pkg/rsstorage/servers/file/server.go
@@ -141,14 +141,14 @@ func (s *StorageServer) CalculateUsage() (types.Usage, error) {
 // diskUsage will walk the specified path in a filesystem and
 // aggregate the size of the contained files.
 func diskUsage(duPath string, cacheTimeout, walkTimeout time.Duration) (size datasize.ByteSize, err error) {
-	stop := make(chan bool)
+	stop := make(chan struct{})
 	defer close(stop)
 
 	sizeChan := make(chan datasize.ByteSize)
 	// errChan should have a buffer of two items to prevent deadlock between `<-stop` and `errChan<-err`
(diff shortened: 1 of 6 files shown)


pkg/rsstorage/servers/file/server.go
```diff
@@ -127,28 +139,59 @@ func (s *StorageServer) CalculateUsage() (types.Usage, error) {
...
+			start = time.Now()
+		// Success case error will return `nil`
+		case err = <-errChan:
+			return
+		default:
+			time.Sleep(walkCheckTime)
+		}
+
+		if time.Now().Sub(start) > walkTimeout {
+			return 0, walktimeoutErr
```
jonyoder, line 192: If we return due to a `cacheTimoutTimer` tick or `walkTimeout`, should we force the filepath walk goroutine to exit?

jonyoder (commented):
Looks like a great idea! Two questions about error reporting and aborting the new goroutine.

Also, if possible, we should add test coverage for:
* The new walk timeout
* The infinite loop reporting

pkg/rsstorage/servers/file/server_test.go
```diff
@@ -712,6 +714,62 @@ func (s *FileStorageServerSuite) TestUsageTimeout(c *check.C) {
...
+
+		defer os.Remove(f.Name())
+	}
+
+	sz, err := diskUsage("testdata", time.Minute, time.Minute)
+	c.Assert(err, check.IsNil)
+	c.Check(uint64(sz), check.Equals, uint64(expectedSize))
+}
+
+func (s *FileStorageServerSuite) TestDiskUsageLoop(c *check.C) {
```
tylfin, line 736: @jonyoder I read that creating a symlink would recreate the loop issue, but I wasn't able to reproduce it here. Any ideas?

jonyoder: I'm not familiar with the issue, so I'm afraid not. Was this related to a bug that is documented elsewhere?

tylfin: Oh, maybe this was an older Go bug, I see something [for Windows ](https://github.com/golang/go/issues/17540) but it looks like it's been closed for a while.

I'm still not sure exactly what the customers are experiencing here, all I know is that this area of the code ends up hanging forever preventing RSPM boot up.

tylfin:
Should be good to go

pkg/rsstorage/servers/file/server.go
```diff
@@ -127,28 +139,64 @@ func (s *StorageServer) CalculateUsage() (types.Usage, error) {
...
 		if err != nil {
-			return err
+			errChan <- err
 		}
 
-		if time.Now().After(timeout) {
-			return errors.New("timeout in diskUsage")
-		}
+		close(sizeChan)
+		close(errChan)
```
jonyoder, line 174: This won't change the behavior now, but you might consider using `defer` for these earlier in case we ever add any short-circuit returns.

pkg/rsstorage/servers/file/server.go
```diff
@@ -127,28 +139,64 @@ func (s *StorageServer) CalculateUsage() (types.Usage, error) {
...
 
-		return nil
-	})
-	size = *sizep
+	start := time.Now()
 
-	return
+	for {
+		if walkTimeout > 0 && time.Now().Sub(start) > walkTimeout {
+			return 0, walktimeoutErr
```
jonyoder, line 184: I may be missing something here, but could we simply establish a `walkTimeoutTimer` above and add it to the `select` below to eliminate this conditional?

pkg/rsstorage/servers/file/server.go
```diff
@@ -286,32 +334,83 @@ func (s *StorageServer) Remove(dir, address string) error {
...
+		if err != nil {
+			errChan <- err
 		}
-		if !info.IsDir() {
-			relPath, err := filepath.Rel(s.dir, path)
-			if err != nil {
-				return err
+
+		close(itemChan)
+		close(errChan)
```
jonyoder, line 387: Maybe add these to a `defer`, as well.

jonyoder (approved):
Looks good! I added a few suggestions, but nothing blocking.

pkg/rsstorage/servers/file/server.go
tylfin: Oh there's a `.Reset()` function isn't there? I was trying to avoid the `time.After(...)` memory leak, but totally forgot about this functionality.

pkg/rsstorage/servers/file/server.go
```diff
@@ -127,28 +139,64 @@ func (s *StorageServer) CalculateUsage() (types.Usage, error) {
 
 // diskUsage will walk the specified path in a filesystem and
 // aggregate the size of the contained files.
-func diskUsage(duPath string, cacheTimeout time.Duration) (size datasize.ByteSize, err error) {
-	timeout := time.Now().Add(cacheTimeout)
-	sizep := &size
+func diskUsage(duPath string, cacheTimeout, walkTimeout time.Duration) (size datasize.ByteSize, err error) {
+	stop := make(chan bool)
```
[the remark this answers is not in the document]
jonyoder: That's a good idea. I like using `chan struct{}`, too, when possible. In RSPM, we often use `chan bool` since we need the stop channel to also signal when a service has been successfully stopped. For example, `stop <- true` signals the service to stop, and the the service sends a `stop <- true` signal again when it has shut down.

jonyoder (approved):
:100:

PR #62, opened 2022-02-28
Add enhanced terminal logging to rslog

eduardocfalcao:
Adding capability to have the enhanced terminal logging as `logrus` has.  

Now we can use the function `rslog.UseTerminalLogger(rslog.InfoLevel)`. This will make the default logger in the `rslog` package, have the output to the `stderr` (which will display on the terminal), with colored text format. improving the log readability for CLI applications.

### DIFF
--- a/examples/cmd/testapp/cmd/log.go
+++ b/examples/cmd/testapp/cmd/log.go
@@ -36,11 +36,11 @@ var LogCmd = &cobra.Command{
 
 var TerminalLogCmd = &cobra.Command{
 	Use:     "terminal-log",
-	Short:   "Command to log some information",
+	Short:   "Command to log some information using the terminal logging style",
 	Example: "",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		rslog.UseTerminalLogger(rslog.InfoLevel)
-		rslog.Infof(message)
+		rslog.WithField("some-field", "some-value").Infof(message)
 		return nil
 	},
 }
--- a/examples/cmd/testapp/cmd/log.go
+++ b/examples/cmd/testapp/cmd/log.go
@@ -20,7 +20,7 @@ func init() {
 	LogCmd.Flags().StringVar(&message, "message", "default message", "The message to log.")
 	CaptureLogCmd.Flags().BoolVar(&withMetadata, "withMetadata", false, "The option to turn on or off the metadata in the capturing logger")
 
-	RootCmd.AddCommand(LogCmd, CaptureLogCmd, CompositeLogCmd)
+	RootCmd.AddCommand(LogCmd, TerminalLogCmd, CaptureLogCmd, CompositeLogCmd)
 }
 
 var LogCmd = &cobra.Command{
@@ -29,6 +29,17 @@ var LogCmd = &cobra.Command{
 	Example: "",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		// Log an info-level message
+		rslog.Infof(message)
+		return nil
+	},
+}
+
+var TerminalLogCmd = &cobra.Command{
+	Use:     "terminal-log",
+	Short:   "Command to log some information",
+	Example: "",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		rslog.UseTerminalLogger(rslog.InfoLevel)
 		rslog.Infof(message)
 		return nil
 	},
--- a/pkg/rslog/logger_impl.go
+++ b/pkg/rslog/logger_impl.go
@@ -18,6 +18,8 @@ var DefaultLoggerFactory LoggerFactory
 type LoggerFactoryImpl struct{}
 
 func (f *LoggerFactoryImpl) DefaultLogger() Logger {
+	// Ignoring error because it only can return error from the builder,
+	// and the builder in use here with the LogOutputStdout option, doesn't return error
 	lgr, _ := NewLoggerImpl(LoggerOptionsImpl{
 		Output: []OutputDest{
 			{
@@ -32,11 +34,15 @@ func (f *LoggerFactoryImpl) DefaultLogger() Logger {
 	return lgr
 }
 
-func (f *LoggerFactoryImpl) TerminalLogger(l LogLevel) Logger {
+type TerminalLoggerFactory struct {
+	LogLevel LogLevel
+}
+
+func (f *TerminalLoggerFactory) DefaultLogger() Logger {
 	// Ignoring error because it only can return error from the builder, and the builder in use here, doesn't return error
 	lgr, _ := NewLoggerImpl(LoggerOptionsImpl{
 		Format: TextFormat,
-		Level:  l,
+		Level:  f.LogLevel,
 	}, StderrOutputBuilder{})
 	return lgr
 }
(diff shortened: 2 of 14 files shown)


examples/cmd/testapp/cmd/log.go
```diff
@@ -34,6 +34,17 @@ var LogCmd = &cobra.Command{
 	},
 }
 
+var TerminalLogCmd = &cobra.Command{
+	Use:     "terminal-log",
+	Short:   "Command to log some information",
```
zackverham, line 39: I'm not sure if we actually use / care about these command instantiation params - if we do, it could be good to add some more description of what the terminal log command is. I needed the PR description to understand what this change is doing, I could see it being confusing without that context.

pkg/rslog/logger_impl.go
```diff
@@ -32,14 +32,16 @@ func (f *LoggerFactoryImpl) DefaultLogger() Logger {
 	return lgr
 }
 
-type LoggerImpl struct {
-	*logrus.Logger
+func (f *LoggerFactoryImpl) TerminalLogger(l LogLevel) Logger {
+	lgr, _ := NewLoggerImpl(LoggerOptionsImpl{
```
zackverham, line 36: just for my own understanding - why do we not care about the possible `err` here?

pkg/rslog/output_builder_test.go
```diff
@@ -4,75 +4,109 @@ package rslog
...
 	"github.com/stretchr/testify/mock"
 )
 
+type ioWriterMock struct {
+	mock.Mock
+}
+
+func (m *ioWriterMock) Write(p []byte) (int, error) {
+	args := m.Called(p)
+	return args.Int(0), args.Error(1)
```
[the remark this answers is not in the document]
marcosnav: this won't return error unless that is specified with something like `wrtMock.On("Write").Return(1, someError)`. 

`args.Error(1)` is some kind of args-casting method that also allows `nil` :)

pkg/rslog/logger_impl.go
marcosnav: Same question, I assume that because this time it is using a StderrOutputBuilder, it isn't possible to error... but want to be sure. Also, if that's the case, having a comment like `// Ignoring error because ...` will be good for future reference :)

marcosnav (approved):
Looking good :)

pkg/rslog/logger_impl.go
gbrlsnchs: Not the author but I think it is like that for two reasons:
1. It is a counterpart for `DefaultLogger`, and thus conforms with that, which also doesn't return any errors
2. Taking a look at the implementation from `NewLoggerImpl`, the error returned is from a call to `(*StderrOutputBuilder).Build`, which never returns any errors, so it's all good

pkg/rslog/output_builder.go
```diff
@@ -89,14 +105,29 @@ func NewOutputLogBuilder(logCategory LogCategory, defaultFile string) OutputBuil
...
+		return b.fileSystem.Stderr(), nil
 	case LogOutputStdout:
-		return os.Stdout, nil
+		return b.fileSystem.Stdout(), nil
 	case LogOutputStderr:
-		return os.Stderr, nil
+		return b.fileSystem.Stderr(), nil
 	case LogOutputFile:
 		return b.createLogFile(logFilePath)
 	default:
```
gbrlsnchs, line 133: Small nit, but this block could be rewritten as:
```golang
switch output {
case LogOutputDefault, LogOutputStderr:
    return b.fileSystem.Stderr(), nil
case LogOutputStdout:
    return b.fileSystem.Stdout(), nil
// ...
}
```

pkg/rslog/output_builder.go
```diff
@@ -89,14 +105,29 @@ func NewOutputLogBuilder(logCategory LogCategory, defaultFile string) OutputBuil
...
+	for _, output := range outputs {
+		w, err := b.build(output.Output, output.Filepath)
+		if err != nil {
+			return nil, err
+		}
+
+		writers = append(writers, w)
+	}
+
+	return io.MultiWriter(writers...), nil
```
gbrlsnchs, line 120: Oh I see, this one has support for multi writer at implementation level. Cool!

gbrlsnchs (approved):

gbrlsnchs:
LGTM! Some questions and nits, you can ignore those if you want. :smiling_face_with_tear:

examples/cmd/testapp/cmd/log.go
eduardocfalcao: Even this being just a sample application, yeah you're right, I will put a better description here.

pkg/rslog/logger_impl.go
eduardocfalcao: Exactly, I'm ignoring the error here because only the builder could return an error, and the `StderrOutputBuilder` doesn't return errors. But thinking better, It would be good to check the errors because we would be already covered in case of any change to the `NewLoggerImpl`. WDYT?

gbrlsnchs: @eduardocfalcao I think if we do that for `TerminalLogger`, maybe we would want to do that for `DefaultLogger` as well. Even because `DefaultLogger` seems able to return an error, and I think we're currently ignoring it. I understand that would be a lot less convenient so I would agree of only doing so for `TerminalLogger`.

pkg/rslog/output_builder_test.go
```diff
@@ -89,11 +123,20 @@ func (s *LoggerSuite) TestBuildOpenFileError() {
...
+
+	result, err := builder.Build(
+		OutputDest{
+			Output:   LogOutputFile,
+			Filepath: "/custom/test.log",
+		})
 
-	result, err := builder.Build(LogOutputFile, "/custom/test.log")
 	s.Nil(err)
-	s.Equal(result, defaultLogPath)
```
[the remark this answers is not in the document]
eduardocfalcao: Yeah, the thing is, now the builder returns an `io.multiWriter`. But, unfortunately, this struct is not public on the `io` package, I can't check if the result is that, and I can't do a contains this `writerMock` on that slice of `writers`, for example. 

So the solution to that:
- As it is a struct, calling the `Write` function, it will call the `Write` of all the writers inside it; 
- I'm filling it with mock writers and checking If the writer I want is receiving a call to the `Writer` function when I call this function of the `multiWriter` returned.

About your second question, what makes a difference here is the different options and mocked stuff on the `mockFileSystem.` If the parameters passed on the mocking statement don't match the ones in the testing phase, it will make the test fail.

pkg/rslog/output_builder_test.go
```diff
@@ -216,6 +299,56 @@ func (s *LoggerSuite) TestLogOutputTypeUnmarshalText() {
...
+		{
+			Output: LogOutputStderr,
+		},
+	}
+
+	result, err := builder.Build(outputDest...)
+	s.Nil(err)
+
+	content := []byte("mock-content")
+	fileExpectedWriter.On("Write", content).Return(len(content), nil)
```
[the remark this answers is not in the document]
eduardocfalcao: The `On` calls prepare the mock object to receive the function calls and know how it will behave for that situation.
In the `AssertExpectations` function, try to ensure that every expectation you've prepared was triggered. 

So, If you want to have your mock object answering a function call, you will only use the `On` method.
If your test phase doesn't reach the mock call you prepared, there is no problem; it won't take the test to fail.

If you want to fail if the mock call was not triggered, you need to use the `AssertExpectations`.

pkg/rslog/logger_impl.go
eduardocfalcao: Yeah, we only worry about the errors when creating outside the `rslog` package. By now, I think I will only add comments to explain why we are ignoring the errors here.

It can be changed, but I'm moving the discussion for later.

zackverham (approved):
The remaining comments seem like they're up to you, I'm happy either way! 

Everything looks good to me.

marcosnav (approved):
✨

marcosnav (approved):

pkg/rslog/logger_impl.go
```diff
@@ -218,11 +219,24 @@ func UpdateDefaultLogger(options LoggerOptionsImpl, outputBuilder OutputBuilder)
...
 }
 
+// UseTerminalLogger sets the DefaultLoggerFactory variable with a TerminalLoggerFactory instance.
+// It makes the default logger have its output to STDERR with enhanced text formatting.
+// Make sure to call it before any call to DefaultLogger or package logging function
+// because the logger instance will be created for the first time that the DefaultLogger function
+// is called.
+func UseTerminalLogger(l LogLevel) {
+	DefaultLoggerFactory = &TerminalLoggerFactory{l}
+}
```
gbrlsnchs, line 238: Considering this will only be used by CLI programs, I think this is a fine solution. :smiley:

gbrlsnchs (approved):

PR #64, opened 2022-03-15
Add IP Reporter interface

nodivbyzero:
This PR adds:

* IPReporter interface
* IP cache interface that will ensure that the same IP is always returned unless the IP of the machine has actually changed and the original IP is no longer valid.
* Unit-tests for IPReporter and IPCache


pkg/rsnotify/listener/ipreporter.go
```diff
@@ -0,0 +1,62 @@
...
+			return false
+		}
+
+		for _, addr := range addrs {
+			switch v := addr.(type) {
+			case *net.IPNet:
+				if v.Contains(ip) {
+					return true
+				}
+			case *net.IPAddr:
```
jonyoder, line 53: Can you add a comment that mentions that `net.IPAddr` is sometimes returned by Windows? I don't see any case on BSD, Linux, or Darwin that returns one of these.

jonyoder (commented):
Looks great to me! A couple of requests for small changes, comments, and tests, but I think this is looking good.

pkg/rsnotify/listener/ipreporter.go
```diff
@@ -0,0 +1,62 @@
...
+	return c.cachedValue
+}
+
+func (p *IPCache) stillValidIP(IP string) bool {
+	var ip net.IP
+	if ip = net.ParseIP(IP); ip == nil {
+		return false
+	}
+
+	interfaces, err := net.Interfaces()
```
jonyoder, line 36: Instead of enumerating interfaces here and then following by enumerating the `inter.Addrs()`, I think we can use a single iteration of `net.InterfaceAddrs()` that gives us all the IP addresses at once.

jonyoder:
@nodivbyzero I think your choices of locations and file names are fine!

pkg/rsnotify/listener/ipreporter.go
nodivbyzero: According to [interface_linux.go](https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/net/interface_linux.go;l=239) it's possible to see IPAddr on Linux.

jonyoder (approved):
Looks good! I'm working on #65 which will make it easier to preserve the `New*` method signatures in this library. I'll wait to merge #65 until this is merged, and then we can cut a new release.

jonyoder (approved):

PR #66, opened 2022-03-22
Add examples for cache, queue, notifications, storage

jonyoder:
* Adds an example application that demonstrates how to use the logging, cache, queue, storage, and notifications components together to render markdown to HTML.
* Updates the LICENSE.md list.
* Updates module READMEs to properly reference example applications.

This isn't intended to be a comprehensive example. I'm hoping to follow up later with more complete examples. For example:
* How to use S3 storage
* Optional in-memory caching
* Object storage and caching
* Queue groups
* Leader election
* Leader scheduled and persistent tasks


jonyoder (commented):
I self reviewed this since it's an example. I'll be merging it after the cache, queue, and election port merges.

PR #69, opened 2022-03-23
Fix dependency versions for rsnotify

jonyoder:


examples/go.mod
```diff
@@ -11,22 +11,31 @@ require (
...
+	github.com/spf13/cobra v1.2.1
+	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c
+	gorm.io/driver/sqlite v1.3.1
+	gorm.io/gorm v1.23.2
+)
+
+// Require platform-lib dependencies. Note that the versions don't matter here
+// since all get replaced by the `replace` directive below. The versions here
+// should match some valid version to make the IDE happy, but they're ignored
+// by Go when compiling.
```
jonyoder, line 23: Because of the `replace` directives below, none of the changes here really make a difference. The main thing is that I added some comments to clarify what's happening.

pkg/rsnotify/listeners/local/go.mod
```diff
@@ -5,12 +5,10 @@ go 1.17
 require (
 	github.com/fortytw2/leaktest v1.3.0
 	github.com/google/uuid v1.1.2
-	github.com/rstudio/platform-lib/pkg/rsnotify v1.2.0
+	github.com/rstudio/platform-lib/pkg/rsnotify v1.4.0
 	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f
 )
 
-replace github.com/rstudio/platform-lib/pkg/rsnotify => ../../
```
jonyoder, line 12: Here, and in the other listener implementations, I removed the `replace` directives. This is better, I think, since it will help us avoid false positive builds. Builds will now fail if we fail to update a listener to use the correct `rsnotify` dependency version.

nodivbyzero (approved):
Thank you for updating those versions.

PR #78, opened 2022-03-28
Change Type() signature to return interface and error

nodivbyzero:
This PR changes TypeMatcher's Type() signature to return interface and error.

This change is needed to address an [issue](https://github.com/rstudio/package-manager/issues/7714 ) where the offline node is receiving messages for queue work (work ready, work complete), but it is not configured to handle those message types. This error is innocuous, but could be alarming to customers.

### DIFF
--- a/pkg/rsnotify/listener/genericmatcher.go
+++ b/pkg/rsnotify/listener/genericmatcher.go
@@ -1,5 +1,7 @@
 package listener
 
+// Copyright (C) 2022 by RStudio, PBC.
+
 import (
 	"errors"
 	"fmt"
--- a/pkg/rsnotify/listener/genericmatcher.go
+++ b/pkg/rsnotify/listener/genericmatcher.go
@@ -1,0 +1,27 @@
+package listener
+
+// Copyright (C) 2022 by RStudio, PBC.
+
+type GenericMatcher struct {
+	field string
+	types map[uint8]interface{}
+}
+
+func (m *GenericMatcher) Field() string {
+	return m.field
+}
+
+func (m *GenericMatcher) Type(notifyType uint8) (interface{}, error) {
+	return m.types[notifyType], nil
+}
+
+func (m *GenericMatcher) Register(notifyType uint8, dataType interface{}) {
+	m.types[notifyType] = dataType
+}
+
+func NewMatcher(field string) *GenericMatcher {
+	return &GenericMatcher{
+		field: field,
+		types: make(map[uint8]interface{}),
+	}
+}
--- a/pkg/rsnotify/listener/genericmatcher.go
+++ b/pkg/rsnotify/listener/genericmatcher.go
@@ -1,6 +1,12 @@
 package listener
 
+import (
+	"errors"
+)
+
 // Copyright (C) 2022 by RStudio, PBC.
+
+var MissingTypeError = errors.New("MissingType error")
 
 type GenericMatcher struct {
 	field string
@@ -12,7 +18,10 @@ func (m *GenericMatcher) Field() string {
 }
 
 func (m *GenericMatcher) Type(notifyType uint8) (interface{}, error) {
-	return m.types[notifyType], nil
+	if t, ok := m.types[notifyType]; ok {
+		return t, nil
+	}
+	return nil, MissingTypeError
 }
 
 func (m *GenericMatcher) Register(notifyType uint8, dataType interface{}) {
--- a/pkg/rsnotify/listener/genericmatcher.go
+++ b/pkg/rsnotify/listener/genericmatcher.go
@@ -2,6 +2,7 @@ package listener
 
 import (
 	"errors"
+	"fmt"
 )
 
 // Copyright (C) 2022 by RStudio, PBC.
@@ -21,7 +22,7 @@ func (m *GenericMatcher) Type(notifyType uint8) (interface{}, error) {
 	if t, ok := m.types[notifyType]; ok {
 		return t, nil
 	}
-	return nil, MissingTypeError
+	return nil, fmt.Errorf("no matcher type found for %d: %w", notifyType, MissingTypeError)
 }
 
 func (m *GenericMatcher) Register(notifyType uint8, dataType interface{}) {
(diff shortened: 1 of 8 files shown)


pkg/rsnotify/listener/genericmatcher_test.go
```diff
@@ -0,0 +1,59 @@
...
+	type TestNotification struct{}
+	m := &GenericMatcher{
+		field: "test",
+		types: make(map[uint8]interface{}),
+	}
+	m.Register(testNotificationType, TestNotification{})
+	c.Assert(m.types[testNotificationType], check.NotNil)
+
+	t, _ := m.Type(testNotificationType)
+	c.Assert(t, check.NotNil)
```
jonyoder, line 58: Can we check the `err` being returned to prove it is nil?

Also, can we add a test where we pass an unknown type and verify that it returns `nil` both for the type and the error?

pkg/rsnotify/listener/listener_test.go
```diff
@@ -3,13 +3,9 @@ package listener
...
-
 	"gopkg.in/check.v1"
 )
 
-type NotifySuite struct{}
-
-var _ = check.Suite(&NotifySuite{})
+type ListenerSuite struct{}
 
-func TestPackage(t *testing.T) { check.TestingT(t) }
```
jonyoder, line 15: Removing this is causing the build to fail. You need one call like this per package. You can either revert this one or move it to `genericmatcher_test.go`.

pkg/rsnotify/listeners/postgrespgx/pgxlistener.go
```diff
@@ -224,13 +224,17 @@ func (l *PgxListener) notify(n *pgconn.Notification, errs chan error, items chan
 		errs <- fmt.Errorf("error unmarshalling message data type: %s", err)
 		return
 	}
-	if l.matcher.Type(dataType) == nil {
+	t, err := l.matcher.Type(dataType)
+	if err != nil {
+		return
+	}
+	if t == nil {
```
jonyoder, line 231: If you want to be able to ignore errors, you'll need to remove this block. I think we probably want this behavior in the generic matcher. Then, using the generic matcher will return this error when no type is found, but we can create a custom matcher for offline mode that returns no error.

pkg/rsnotify/listeners/postgrespq/pqlistener.go
```diff
@@ -183,13 +183,17 @@ func (l *PqListener) notify(n *pq.Notification, errs chan error, items chan list
 		errs <- fmt.Errorf("error unmarshalling message data type: %s", err)
 		return
 	}
-	if l.matcher.Type(dataType) == nil {
+	t, err := l.matcher.Type(dataType)
+	if err != nil {
+		return
+	}
+	if t == nil {
```
jonyoder, line 190: See comment on pgxlistener.go.

jonyoder (commented):
This looks like it's almost ready. One concern about error handling, and I noticed a test-wiring issue.

pkg/rsnotify/listeners/postgrespgx/pgxlistener.go
nodivbyzero: oops. 
I did the opposite: genericMatcher return no error (no changes in the GenericMatcher) and OfflineMatcher return `MissingType` error.

What kind of interface should OfflineMatcher return in case of missing type?

jonyoder: I was thinking that the generic matcher would return the error on a missing type, while the offline matcher would simply return a nil error. Then, in the listener implementations, if we get back a nil value and no error, we would simply return with no error.

pkg/rsnotify/listener/genericmatcher.go
```diff
@@ -12,7 +18,10 @@ func (m *GenericMatcher) Field() string {
 }
 
 func (m *GenericMatcher) Type(notifyType uint8) (interface{}, error) {
-	return m.types[notifyType], nil
+	if t, ok := m.types[notifyType]; ok {
+		return t, nil
+	}
+	return nil, MissingTypeError
```
jonyoder, line 24: I think this is a good case for error wrapping:

```go
return fmt.Errorf("no matcher type found for %d: %w", notifyType, MissingTypeError)
```

pkg/rsnotify/listeners/postgrespgx/pgxlistener.go
```diff
@@ -226,10 +226,10 @@ func (l *PgxListener) notify(n *pgconn.Notification, errs chan error, items chan
 	}
 	t, err := l.matcher.Type(dataType)
 	if err != nil {
+		errs <- fmt.Errorf("no matcher type found for %d", dataType)
```
jonyoder, line 229: Here, and also in the pqlistener, if we switch to error wrapping, we can simply pass the error here:
```
errs <- err
```

pkg/rsqueue/groups/grouprunner.go
```diff
@@ -41,7 +42,10 @@ func (m *GenericMatcher) Field() string {
 }
 
 func (m *GenericMatcher) Type(workType uint64) (interface{}, error) {
-	return m.types[workType], nil
+	if t, ok := m.types[workType]; ok {
+		return t, nil
+	}
+	return nil, MissingTypeError
```
jonyoder, line 48: We can wrap this one, too, and then simply pass the error below as suggested above.

jonyoder (commented):
Just one more suggestion about wrapping errors to make them easier to pass later. After that, I think this is good to merge.

pkg/rsnotify/listener/genericmatcher.go
```diff
@@ -0,0 +1,37 @@
+package listener
+
```
jonyoder, line 2: CI is failing since there is no copyright notice on line 3 here.

jonyoder (approved):
When the copyright notice is fixed and CI passes, let's merge this!

jonyoder:
FYI, you can catch these issues before pushing if you run:
```
just vet
just lint
```

PR #83, opened 2022-04-11
Do not create logging dirs, default log directories should already exist and be created via product install

marcosnav:
Do not create logging dirs, default log directories should already exist and be created via product install

### DIFF
--- a/pkg/rslog/output_builder.go
+++ b/pkg/rslog/output_builder.go
@@ -127,24 +127,35 @@ func (b outputBuilder) build(output LogOutputType, logFilePath string) (io.Write
 	case LogOutputStdout:
 		return b.fileSystem.Stdout(), nil
 	case LogOutputFile:
-		return b.createLogFile(logFilePath)
+		return b.resolveLogFile(logFilePath)
 	default:
 		err := fmt.Errorf("The output %q provided for the logging output configuration field isn't supported.", output)
 		return nil, err
 	}
 }
 
-func (b outputBuilder) createLoggingDir(logFilePath string) {
+func (b outputBuilder) resolveLogFile(logFilePath string) (io.Writer, error) {
 	lgr := DefaultLogger()
+	if strings.TrimSpace(logFilePath) == "" {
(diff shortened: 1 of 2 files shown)


pkg/rslog/output_builder.go
```diff
@@ -127,24 +127,35 @@ func (b outputBuilder) build(output LogOutputType, logFilePath string) (io.Write
...
 		err := fmt.Errorf("The output %q provided for the logging output configuration field isn't supported.", output)
 		return nil, err
 	}
 }
 
-func (b outputBuilder) createLoggingDir(logFilePath string) {
+func (b outputBuilder) resolveLogFile(logFilePath string) (io.Writer, error) {
 	lgr := DefaultLogger()
+	if strings.TrimSpace(logFilePath) == "" {
+		lgr.Infof("Logging output is set to FILE, but no path was provided.")
```
zackverham, line 140: just a thought - L154 is also returning a b.fallbackOutput but logging an Error - should this be an error log as well? Or possibly a warn?

marcosnav: good point, yeah this should be definitely a warn, I'll update 👍

zackverham: hmm, I see - that's just a direct copy from createLogFile - so maybe fine to leave as-is then. 

I'm fine with leaving it - just a thought!

zackverham (approved):
makes sense to me!

Do we know that the idea that default log directories should exist from the install is true across products? I.e. if this is the company-wide product lib just want to make sure that we're not encoding a Connect-specific assumption here.

marcosnav:
@zackverham package-manager also creates logging dirs on post install scripts, AFAIK that's the only pro product to integrate this lib.

https://github.com/rstudio/package-manager/blob/2d04a1d4cc970c062d22fff17563c51565b2794d/packaging/debian-control/postinst.in#L53-L60

Still I'll wait for @jonyoder comments/thoughts 🙂

jonyoder (approved):
This looks fine to me.

PR #86, opened 2022-05-02
Docker images with arm64 support

mcbex:
I am still working on the image for the dev container but the testing process is slow so I will make a separate PR once that is ready.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -33,6 +33,8 @@ ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
 ENV DEBIAN_FRONTEND=noninteractive
 
+ARG TARGETARCH
+
 # Upgrade pip
 RUN pip3 install --upgrade pip
 
@@ -51,11 +53,10 @@ RUN chmod +x /usr/local/bin/docker-compose
 
 # Install Go. Keep in sync with other Dockerfiles.
 ENV GOLANG_VERSION 1.18
-RUN ARCH=$(if [ $(uname -m) = "aarch64" ]; then echo "arm64"; else echo "amd64"; fi) \
-    && GOLANG_DOWNLOAD_SHA256=$(if [ $(uname -m) = "aarch64" ]; \
+RUN GOLANG_DOWNLOAD_SHA256=$(if [ $TARGETARCH = "arm64" ]; \
       then echo "7ac7b396a691e588c5fb57687759e6c4db84a2a3bbebb0765f4b38e5b1c5b00e"; \
       else echo "e85278e98f57cdb150fe8409e6e5df5343ecb13cebf03a5d5ff12bd55a80264f"; fi) \
-    && curl -fsSL "https://dl.google.com/go/go${GOLANG_VERSION}.linux-${ARCH}.tar.gz" -o golang.tar.gz \
+    && curl -fsSL "https://dl.google.com/go/go${GOLANG_VERSION}.linux-${TARGETARCH}.tar.gz" -o golang.tar.gz \
     && echo "$GOLANG_DOWNLOAD_SHA256  golang.tar.gz" | sha256sum -c - \
     && tar -C /usr/local -xzf golang.tar.gz \
     && rm golang.tar.gz
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -37,8 +37,8 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN pip3 install --upgrade pip
 
 # Install AWS CLI v2
-RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
-    unzip awscliv2.zip && \
+RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m).zip" -o "awscliv2.zip" && \
+    unzip -q awscliv2.zip && \
     ./aws/install && \
     rm -rf awscliv2.zip aws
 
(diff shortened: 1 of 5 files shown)


mcbex:
The build error may be related to this issue: https://github.com/golang/go/issues/51253

mcbex:
Also need to look more closely at these issues which were brought to my attention by Aron: 

https://github.com/actions/checkout/issues/760
https://github.com/actions/checkout/issues/766

docker/bionic/Dockerfile
```diff
@@ -51,8 +51,11 @@ RUN chmod +x /usr/local/bin/docker-compose
 
 # Install Go. Keep in sync with other Dockerfiles.
 ENV GOLANG_VERSION 1.18
-ENV GOLANG_DOWNLOAD_SHA256 e85278e98f57cdb150fe8409e6e5df5343ecb13cebf03a5d5ff12bd55a80264f
-RUN curl -fsSL "https://dl.google.com/go/go${GOLANG_VERSION}.linux-amd64.tar.gz" -o golang.tar.gz \
+RUN ARCH=$(if [ $(uname -m) = "aarch64" ]; then echo "arm64"; else echo "amd64"; fi) \
```
jonyoder, line 54: Here, and above, I wonder if we should consider simply using an `ARG ARCH=amd64` and optionally passing the architecture to the build. For example, see the Launcher dockerfile: https://github.com/rstudio/launcher/blob/main/jenkins/docker/Dockerfile.bionic#L1.

Using an `ARG` would keep the Dockerfile a bit cleaner and let us handle the shell logic elsewhere.

mcbex: I think the issue would be that when we install awscli and Go the URLs contain the architecture but it's not consistent: aarch64 and arm64 vs x86_64 and amd64. So passing in a single variable would not remove the need to resolve those URLs dynamically.... unless you're suggesting the part where we build the URLs should move out of the Dockerfile? 

A default architecture hard-coded in the image would also mean that anyone not on an amd64 machine has to remember to pass the parameter every time. That really only affects me at the moment but it feels kind of clunky to require from users. 

Maybe another solution is to have two Dockerfiles and separate build targets for each architecture?

mcbex:
I'm not too familiar with VS code or developing in containers but I had to update how the dev container was installing Go modules to be the same as the other Dockerfile to get the image to build. I installed VS code and was able to connect to the container afterwards. 

Happy to discuss the overall approach more as well.

bdeitte:
I'm not sure if this will help here, but I saw it come up in a ticket from @shepherdjerred for RSPM.  https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

.devcontainer/Dockerfile
```diff
@@ -37,8 +37,8 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN pip3 install --upgrade pip
 
 # Install AWS CLI v2
-RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
-    unzip awscliv2.zip && \
+RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m).zip" -o "awscliv2.zip" && \
```
shepherdjerred, line 40: Another way to do this is with `TARGETARCH` instead of shelling out: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

docker/bionic/Dockerfile
shepherdjerred: I comment this above before I saw this comment. Docker BuildKit will set `TARGETARCH` to the architecture you're building for, so you won't have to pass any extra parameters. The info is pretty far down the page:

https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

```
BUILDPLATFORM — matches the current machine. (e.g. linux/amd64)

BUILDOS — os component of BUILDPLATFORM, e.g. linux

BUILDARCH — e.g. amd64, arm64, riscv64

BUILDVARIANT — used to set ARM variant, e.g. v7

TARGETPLATFORM — The value set with --platform flag on build

TARGETOS - OS component from --platform, e.g. linux

TARGETARCH - Architecture from --platform, e.g. arm64

TARGETVARIANT
```

shepherdjerred:
> I'm not sure if this will help here, but I saw it come up in a ticket from @shepherdjerred for RSPM. https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

Thanks for the mention Brian! The solution here looks similar to what I'm doing for RSPM.

docker/bionic/Dockerfile
mcbex: Got it. Thanks for the feedback @jonyoder @shepherdjerred 

I briefly read some of the documentation on BuildKit features but I wasn't sure if using BuildKit was considered portable or not (since it looked pretty new). So good to know others are also looking at this and thinking about it!

shepherdjerred: BuildKit is now about 3 years old: https://docs.docker.com/engine/release-notes/18.09/

As far as I know, it's the best way to make cross-platform images. You are right that it's not completely mature, this issue shows the gaps: https://github.com/moby/moby/issues/40379

I believe that BuildKit is the default builder now, so that at least shows that Docker considers it to be the way forward: https://github.com/docker/cli/pull/3314

mcbex: TIL!

mcbex:
I was able to remove the logic for downloading the Go tarball by using the TARGETARCH env var which simplified the Dockerfile quite a bit but I couldn't think of a better way to set the correct checksum. Other suggestions welcome.

.devcontainer/Dockerfile
jonyoder: Are we able to update these `uname` uses with `TARGETARCH` instead?

jonyoder (commented):
One remaining question! Looks great!

.devcontainer/Dockerfile
```diff
@@ -50,28 +52,36 @@ RUN curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-c
 RUN chmod +x /usr/local/bin/docker-compose
 
 # Install Go. Keep in sync with other Dockerfiles.
-ENV GOLANG_VERSION 1.17.1
-ENV GOLANG_DOWNLOAD_SHA256 dab7d9c34361dc21ec237d584590d72500652e7c909bf082758fb63064fca0ef
-RUN curl -fsSL "https://dl.google.com/go/go${GOLANG_VERSION}.linux-amd64.tar.gz" -o golang.tar.gz \
+ENV GOLANG_VERSION 1.18
+RUN GOLANG_DOWNLOAD_SHA256=$(if [ $TARGETARCH = "arm64" ]; \
+      then echo "7ac7b396a691e588c5fb57687759e6c4db84a2a3bbebb0765f4b38e5b1c5b00e"; \
+      else echo "e85278e98f57cdb150fe8409e6e5df5343ecb13cebf03a5d5ff12bd55a80264f"; fi) \
```
shepherdjerred, line 58: From what I've seen this is (unfortunately) the idiomatic way to do platform-related logic in Dockerfiles. One note though; explicitly check for amd64, don't assume that arm64 and amd64 are the only possible architectures.

.devcontainer/Dockerfile
```diff
@@ -50,28 +52,36 @@ RUN curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-c
...
+    github.com/josharian/impl@latest
+RUN PATH="$PATH:/usr/local/go/bin" GOPATH=/usr/local/gotools go install \
+    github.com/haya14busa/goplay/cmd/goplay@latest
+RUN PATH="$PATH:/usr/local/go/bin" GOPATH=/usr/local/gotools go install \
+    honnef.co/go/tools/cmd/staticcheck@latest
+RUN PATH="$PATH:/usr/local/go/bin" GOPATH=/usr/local/gotools go install \
+    github.com/uudashr/gopkgs/v2/cmd/gopkgs@latest
 RUN PATH="$PATH:/usr/local/go/bin" go install github.com/go-delve/delve/cmd/dlv@latest
-RUN GO111MODULE=on PATH="$PATH:/usr/local/go/bin" GOPATH=/usr/local/gotools go get -v \
+RUN GO111MODULE=on PATH="$PATH:/usr/local/go/bin" GOPATH=/usr/local/gotools go install \
```
shepherdjerred, line 80: Should GO111MODULE still be here? I thought it was removed in 1.18.

shepherdjerred (approved):

jonyoder (approved):
Marking this as approved so I don't hold things up. If you're able to look into the GO111MODULE question and refactor the `unames`, this looks great to me!

jonyoder:
@mcbex should we merge this?

mcbex:
Ugh my apologies I never saw these comments. Going to go over it again and make those changes!

.devcontainer/Dockerfile
mcbex: Unfortunately we can't swap in `TARGETARCH` here because awscli uses `aarch64` and `x86_64` in the URL instead of `arm64` and `amd64`. However I can check the value of `TARGETARCH` and set an env var instead of shelling out if that is preferred.

.devcontainer/Dockerfile
```diff
@@ -1,5 +1,11 @@
 FROM ubuntu:bionic
 
+ARG TARGETARCH
+
+RUN if [ "${TARGETARCH}" != "amd64" ] && [ "${TARGETARCH}" != "arm64" ]; then \
+    echo "${TARGETARCH} is not a supported architecture. Please use linux/amd64 or linux/arm64"; \
+    exit 1; fi
```
mcbex, line 7: I don't love this but realized while testing that waiting until the later build stages to discover a particular platform is unsupported is annoying. I couldn't find a built in mechanism to allowlist architectures but happy to hear better ideas.

mcbex:
@jonyoder Let me know if you see any other changes I should make before we merge

jonyoder (approved):

PR #87, opened 2022-05-03
'-buildvcs=false' for temporary build fix

jonyoder:
Attempting to fix the build issues. This updates the GHA configuration to also use Go 1.18 (used by dockerfile), and uses the `-buildvcs=false` flag where necessary to fix the build errors. 

I think the build errors are due to older versions of Git that don't support Go's VCS stamping, so we could consider updating Git instead.

### DIFF
--- a/.github/workflows/github-actions-ci.yml
+++ b/.github/workflows/github-actions-ci.yml
@@ -9,7 +9,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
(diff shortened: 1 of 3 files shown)


mcbex (approved):
👍

aronatkins:
FYI: There's additional work planned for the 1.18.x bugfix patch. https://github.com/golang/go/issues/51748 I'm holding at 1.17.x until some of this churn settles.

PR #88, opened 2022-06-03
adding entry formatters that enforce UTC time conversion

zackverham:
Creates custom JSON / Text entry formatters which enforce UTC time conversion.

These are basically just wrappers around the logrus formatters which perform an extra time conversion on the entry field. 

Solution reference: 

https://stackoverflow.com/questions/40502309/how-can-i-set-the-logrus-time-to-utc

Fixes: 

https://github.com/rstudio/connect/issues/21579

### DIFF
--- a/pkg/rslog/rslogtest/logger_impl_test.go
+++ b/pkg/rslog/rslogtest/logger_impl_test.go
@@ -4,7 +4,9 @@ package rslogtest
 
 import (
 	"fmt"
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/rstudio/platform-lib/pkg/rslog"
 	"github.com/sirupsen/logrus"
@@ -50,7 +52,7 @@ func (s *LoggerImplTestSuite) TestNewLoggerImpl() {
 	)
 	s.Nil(err)
 	s.Equal(expectedOutput, result.Out)
-	s.IsType(&logrus.TextFormatter{}, result.Formatter)
+	s.IsType(&rslog.UTCTextFormatter{}, result.Formatter)
 
 	result, err = rslog.NewLoggerImpl(
 		rslog.LoggerOptionsImpl{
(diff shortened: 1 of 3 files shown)


pkg/rslog/rslogtest/logger_impl_test.go
```diff
@@ -352,3 +354,101 @@ func (s *LoggerImplTestSuite) TestNewDiscardingLogger() {
...
+
+	expHour, expMin, expSec := expected.Clock()
+	nonUtcHour, nonUtcMin, nonUtcSec := nonUtc.Clock()
+
+	s.Equal(expHour, nonUtcHour)
+	s.Equal(expMin, nonUtcMin)
+	s.Equal(expSec, nonUtcSec)
+
+	// verify shifted time is not in UTC location
+	s.NotEqual(utc.Location(), nonUtc.Location())
```
jonyoder, line 434: You might consider moving lines `359-384` and `409-434` to a helper method and invoking it here like:

```
utc, nonUtc := getTestTimes(c)
```

That would help to clarify exactly what we're testing in these two unit tests. I was about to comment that we're testing the `time` package in these lines, but then I realized that it is useful to prove that we've actually created the times we intended.

jonyoder (approved):
Looks great! Thanks for this change. One suggestion to make the tests a bit more readable.

pkg/rslog/rslogtest/logger_impl_test.go
zackverham: good call - just refactored those lines. 

Does my change look ok in terms of go best practices?

jonyoder: Looks great to me! Thanks.

marcosnav (approved):
Thanks Zack!

PR #92, opened 2022-10-12
Periodicially reassign the PgxListener IP

tylfin:
One idea to fix the `... node list length differs. Store node count..` error customers are seeing.

For example, one way this can happen is:

1. RSPM node A starts and connects to a severless aurora PostgreSQL DB
2. DB returns that node A connection is "1.1.1.1" and node A records this
3. X amount of time passes and the DB rotates servers
4. DB returns that node A connection is "1.1.1.2" and RSPM node believes it is "1.1.1.1"

Since the `PgxListener.ip` is static, this leads to an unrecoverable loop until the listener is restarted.

This fix simply re-evaluates the IP returned from the DB every 5 minutes.

### DIFF
--- a/pkg/rsnotify/listeners/postgrespgx/pgxlistener.go
+++ b/pkg/rsnotify/listeners/postgrespgx/pgxlistener.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"net"
 	"reflect"
+	"sync"
 	"time"
 
 	"github.com/jackc/pgconn"
@@ -63,6 +64,9 @@ type PgxListener struct {
 	ip          string
 	debugLogger listener.DebugLogger
 	ipReporter  listener.IPReporter
+	ipRefresh   time.Duration
+
+	mu sync.RWMutex
 
 	// For caching chunked notifications
 	notifyCache map[string]map[int][]byte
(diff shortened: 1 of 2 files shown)


pkg/rsnotify/listeners/postgrespgx/pgxlistener.go
```diff
@@ -89,6 +90,15 @@ func NewPgxListener(args PgxListenerArgs) *PgxListener {
...
 
 func (l *PgxListener) IP() string {
+	if l.ip != "" {
+		return l.ip
+	}
+
+	l.ip = l.ipReporter.IP()
+	go func() {
+		time.Sleep(5 * time.Minute)
+		l.ip = ""
```
jonyoder, line 100: Do we need a RWMutex for clearing this value (and reading it)?

tylfin: Yeah, I wavered on this, but you could probably get into the situation with an empty string being returned. The "right," way is to probably setup a goroutine with a ticker and an RWMutex.

I'll take a look, I didn't want to get into making a stopper channel.

jonyoder (commented):
This looks great. Is it difficult to add a unit test that (a) proves the refresh happens and (b) proves that the method exits when the context is canceled?

pkg/rsnotify/listeners/postgrespgx/pgxlistener_test.go
```diff
@@ -434,6 +446,9 @@ func (s *PgxNotifySuite) TestNotificationsBlock(c *check.C) {
 	// Wait for test to complete
 	<-done
 
+	// Assert that the IP is being refreshed
```
tylfin, line 449: Note: leaktest ensures that goroutine shuts down properly

PR #93, opened 2022-10-25
Fix unhandled ping requests from the same PgxLeader node

glin:
Follow-up to #92: 

Previously, the leader node was ignoring pings from itself and not updating the in-memory node list. When the leader node's IP changed, it was sweeping itself without re-adding itself under the new IP, leading to errors like `Error verifying cluster integrity: node list length differs. Store node count 1 does not match leader count 0`. 

To reproduce this, you can set up a single-node cluster with an Aurora Serverless v1 Postgres database (more details at https://github.com/rstudio/platform-lib/pull/92).

The proposed solution here is to have the leader handle pings from itself just like follower pings, so it updates itself in the in-memory node list whenever its IP changes. I've been testing this in a cluster that repros the issue, and no longer see the node list length verify errors after the fix.

### DIFF
--- a/pkg/rselection/impls/pgx/leader.go
+++ b/pkg/rselection/impls/pgx/leader.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+
 	"github.com/rstudio/platform-lib/pkg/rselection"
 	"github.com/rstudio/platform-lib/pkg/rselection/electiontypes"
 	"github.com/rstudio/platform-lib/pkg/rsnotify/broadcaster"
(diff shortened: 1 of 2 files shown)


pkg/rselection/impls/pgx/leader.go
```diff
@@ -134,10 +135,10 @@ func (p *PgxLeader) lead(pingTick, sweepTick <-chan time.Time, stop chan bool) {
 	l := p.awb.Subscribe(electiontypes.ClusterMessageTypePingResponse)
 	defer p.awb.Unsubscribe(l)
 
-	// Listen for other ping requests. If an expected ping request is received,
+	// Listen for other ping requests. If an unexpected ping request is received,
 	// we must demote the leader since there's another leader.
-	unexpectedPings := p.awb.Subscribe(electiontypes.ClusterMessageTypePing)
-	defer p.awb.Unsubscribe(unexpectedPings)
+	leaderPings := p.awb.Subscribe(electiontypes.ClusterMessageTypePing)
+	defer p.awb.Unsubscribe(leaderPings)
```
glin, line 141: Renamed this since the leader is expecting pings from itself

pkg/rselection/impls/pgx/leader_test.go
```diff
@@ -363,6 +364,84 @@ func (s *LeaderSuite) TestPingNodes(c *check.C) {
...
 }
 
+func (s *LeaderSuite) TestLeaderPingSelf(c *check.C) {
+	defer leaktest.Check(c)
+
+	channel := c.TestName()
+	// Use a real notifier to send the Ping Request notification.
+	realNotifier := &PgxPgNotifier{pool: s.pool}
+	// Use a fake notifier to record the Ping Response notification.
+	fakeNotifier := &dummyNotifier{}
```
glin, line 374: Writing a unit test for this was really tricky and I couldn't figure out how to use just the real notifier and call `leader.pingNodes()`. But all other tests use the fake notifier + real notifier as well, so I guess there's a reason why we have to use the fake notifier?

jonyoder (approved):
I think this looks great. Thanks for tracking down the issue and working up a fix!

jonyoder:
The one thing I'm still wondering is if we could simplify all of this logic by generating a GUID on on startup and simply using it instead of the IP.

glin:
I think that's a more proper fix than this 5 minute IP refresh too, which is more like a bandaid. But it also feels like a broader change with complications like handling node restarts that we talked about. I can open an issue to move to that eventually.

tylfin (approved):

PR #94, opened 2022-11-21
Enable debug logging regions messages for when legacy debug is removed

marcosnav:
With the removal of the legacy debug package in Connect, we can now have a version of rslog that is responsible for showing messages of the regions enabled.

### DIFF
--- a/pkg/rslog/debug.go
+++ b/pkg/rslog/debug.go
@@ -60,6 +60,7 @@ func InitDebugLogs(regions []ProductRegion) {
 
 	// Reset enabled regions on each call.
 	regionsEnabled = make(map[ProductRegion]bool)
+	lgr := DefaultLogger()
 
 	// match debug log region to list
 	for _, region := range regions {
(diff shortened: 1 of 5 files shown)


pkg/rslog/debug.go
```diff
@@ -60,21 +60,16 @@ func InitDebugLogs(regions []ProductRegion) {
 
 	// Reset enabled regions on each call.
 	regionsEnabled = make(map[ProductRegion]bool)
+	lgr := DefaultLogger()
```
zackverham, line 63: I guess we have to instantiate a new logger here because we don't have the main one instantiated yet, right?

zackverham (approved):

pkg/rslog/debug.go
marcosnav: The `DefaultLogger()` func calls a method that sets a default logger if it doesn't exist yet. https://github.com/rstudio/platform-lib/blob/6838bcc0d0594953823755e3029f9d199d834fed/pkg/rslog/logger_impl.go#L217-L234

jonyoder: Will this have any effect on using a non-default logger factory? I don't think so, but wanted to be sure.

jonyoder: Also, it looks like there's a test failure.

marcosnav: Not a direct impact, but initializing debug logs now will set the default factory if it doesn't exist. That why the test fails, due to the order of init debug regions and then setting the test factory.

marcosnav:
@jonyoder Added a NEWS file. Let me know what do you think https://github.com/rstudio/platform-lib/blob/mnv-debug-logger-msgs/pkg/rslog/NEWS.md

jonyoder (approved):

PR #95, opened 2022-11-30
Buffer logs for default logger

gbrlsnchs:
## Description
This prevents log messages to be printed with undesired outputs/levels while the configuration file of a platform is not read yet.

A new logger implementation is introduced. This implementation buffers log calls --- from the main logger as well from its children loggers --- in order for them to be flushed later on.

The flush is done when the default logger is updated --- signaling that the necessary initial changes have been made, and it is now safe to print messages.

## How to test with Connect
- [x] Build Connect with [this branch](https://github.com/rstudio/connect/tree/gbrlsnchs-buf-logger) (that branch makes Connect use this branch for `rslog`, thus making use of these changes)
- [x] Assert that Connect starts up without panicking
- [x] Assert that no INFO logs get printed when `Logging.ServiceLogLevel` is `error` (make sure no debug log regions are on)
```ini
[Debug]
; Log = rproc
; Log = tfproc
; Log = pyproc
; Log = proc

[Logging]
ServiceLog = stdout
ServiceLogLevel = error
```
- [x] Assert that INFO logs get printed to a file and no entries escape to stdout:
```ini
[Logging]
ServiceLog = service.log
```
- [x] Make sure Connect prints all logs prior to a fatal error (one can force this by adding invalid fields to the config file)
```ini
[Logging]
Foo = bar
```

### DIFF
--- a/pkg/rslog/buf_logger.go
+++ b/pkg/rslog/buf_logger.go
@@ -2,106 +2,103 @@ package rslog
 
 // Copyright (C) 2022 by RStudio, PBC.
 
-import "github.com/sirupsen/logrus"
-
-type logEntry struct {
-	lvl     LogLevel
-	msg     string
-	args    []interface{}
-	wrapper *logrusEntryWrapper
+type BufLogEntry struct {
+	Level   LogLevel
+	Message string
+	Args    []interface{}
+	Logger  loggerImpl
 }
 
-type bufStorage struct {
-	logs     []logEntry
-	wrappers map[*logrusEntryWrapper]struct{}
+type BufStorage struct {
+	Logs []BufLogEntry
 }
 
-type bufLogger struct {
-	wrapper    *logrusEntryWrapper
-	coreLogger loggerImpl
-	storage    *bufStorage
-	flusher    flusher
+type BufLogger struct {
+	CoreLogger loggerImpl
+	Storage    *BufStorage
 }
 
-var _ loggerImpl = new(bufLogger)
+var _ loggerImpl = new(BufLogger)
 
-func newBufLogger(coreLogger *logrus.Logger) *bufLogger {
-	return &bufLogger{
-		storage: &bufStorage{
-			logs:     make([]logEntry, 0),
-			wrappers: make(map[*logrusEntryWrapper]struct{}),
+func NewBufLogger(coreLogger loggerImpl) *BufLogger {
+	return &BufLogger{
+		Storage: &BufStorage{
+			Logs: make([]BufLogEntry, 0),
 		},
-		coreLogger: coreLogger,
+		CoreLogger: coreLogger,
 	}
 }
 
 // child creates and registers a wrapped logger implementation and returns a new buffered logger
 // that will pass down the same storage and fallback fields for its children, if any.
-func (buf *bufLogger) child(wrapper *logrusEntryWrapper) *bufLogger {
-	buf.storage.wrappers[wrapper] = struct{}{}
-	return &bufLogger{
-		wrapper:    wrapper,
-		coreLogger: buf.coreLogger,
-		storage:    buf.storage,
-		flusher:    buf.flusher,
+func (buf *BufLogger) child(coreLogger loggerImpl) *BufLogger {
+	return &BufLogger{
+		CoreLogger: coreLogger,
+		Storage:    buf.Storage,
 	}
 }
 
-func (buf *bufLogger) Tracef(msg string, args ...interface{}) {
-	buf.storage.logs = append(buf.storage.logs, logEntry{
-		lvl:     TraceLevel,
-		msg:     msg,
-		args:    args,
-		wrapper: buf.wrapper,
+func (buf *BufLogger) Tracef(msg string, args ...interface{}) {
+	buf.Storage.Logs = append(buf.Storage.Logs, BufLogEntry{
+		Level:   TraceLevel,
+		Message: msg,
+		Args:    args,
+		Logger:  buf.CoreLogger,
 	})
 }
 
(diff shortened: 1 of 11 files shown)


pkg/rslog/buf_logger.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+func (buf *BufLogger) Tracef(msg string, args ...interface{}) {
+	buf.Storage.Logs = append(buf.Storage.Logs, BufLogEntry{
+		Level:   TraceLevel,
+		Message: msg,
+		Args:    args,
+		Logger:  buf.CoreLogger,
+	})
```
marcosnav, line 49: WDYT about having a `Storage.Add()` method to simplify the append in log methods?

pkg/rslog/logger_impl.go
```diff
@@ -214,6 +286,34 @@ var defaultLogger Logger
...
+func Buffer() {
+	lock := ensureDefaultLoggerReadLock()
+	defer lock.RUnlock()
+
+	if b, ok := defaultLogger.(buffer); ok {
+		b.Buffer()
+	}
+}
+
+func Flush() {
```
marcosnav, line 308: It'll be good to have a description comment for `Flush` as well

pkg/rslog/logger_impl.go
```diff
@@ -214,6 +286,34 @@ var defaultLogger Logger
...
+type buffer interface {
+	Buffer()
+}
+
+type flusher interface {
+	Flush()
+}
+
+// Buffer signals the default logger to buffer calls.
+// Returns a function that flushes the buffer.
```
marcosnav, line 298: I think this comment isn't valid? since `Buffer` doesn't return anything

pkg/rslog/logger_impl.go
```diff
@@ -179,30 +236,45 @@ func (l LoggerImpl) Logf(msg string, args ...interface{}) {
...
+	}
+
+	return wrapper
 }
 
 func (l logrusEntryWrapper) WithFields(fields Fields) Logger {
-	e := l.Entry.WithFields(logrus.Fields(fields))
-	return logrusEntryWrapper{Entry: e}
+	e := l.coreLogger.WithFields(logrus.Fields(fields))
+	wrapper := &logrusEntryWrapper{coreLogger: e}
```
marcosnav, line 271: Why this wrapper isn't setting `CoreLoggerImpl: e, wrappers: l.wrappers`?

pkg/rslog/logger_impl.go
```diff
@@ -140,34 +144,87 @@ func getLevel(level LogLevel) logrus.Level {
...
+
 func (l LoggerImpl) WithField(key string, value interface{}) Logger {
-	e := l.Logger.WithField(key, value)
-	return logrusEntryWrapper{Entry: e}
+	e := l.CoreLogger.WithField(key, value)
+	wrapper := &logrusEntryWrapper{coreLogger: e, CoreLoggerImpl: e, wrappers: l.wrappers}
+	if bufLogger, ok := l.CoreLoggerImpl.(*BufLogger); ok {
+		l.wrappers[wrapper] = struct{}{}
+		wrapper.CoreLoggerImpl = bufLogger.child(wrapper.coreLogger)
+	}
```
marcosnav, line 194: WDYT about having a `wrapBufLoggerChild()` method to consolidate this check blocks across the `WithField(s)` methods?

pkg/rslog/buf_logger.go
marcosnav: Also, there is risk of having a race condition with this multiple `append` calls. You could implement a single mutex lock in the proposed `Storage.Add()` method.

marcosnav (commented):
Looking good! Some comments

pkg/rslog/logger_impl.go
gbrlsnchs: My bad, it's from another implementation I was trying :man_facepalming:

marcosnav (approved):
Good work! We should add an entry in NEWS.md for this changes

jonyoder:
To be more specific, I was thinking of updating the code in https://github.com/rstudio/platform-lib/tree/main/examples/cmd/testlog/cmd to ensure that it demonstrates how logs are buffered and flushed.

gbrlsnchs:
I'll update both NEWS and the example app then!

marcosnav (approved):

gbrlsnchs:
@jonyoder Buffering logs for all commands won't hurt (as long as they're flushed), so I decided to go with that.

I changed the `BufferedLogCmd` to pin the log level to WARN so we can showcase how buffering works (debug INFO logs wouldn't get printed).

jonyoder (approved):

marcosnav:
I'm picking this to be QA'd along with Connect

marcosnav:
## Validation running this with Connect 

**Set Level to error and no INFO logs were printed**
Screen Shot 2022-12-12 at 12 05 53

**Setting a service.log file, all logs are sent to this file**
Screen Shot 2022-12-12 at 12 07 00

**On misconfiguration, logs prior to fatal are able to show up**
```
start-licensed.sh: Running target command ...
time="2022-12-12T18:08:08.679Z" level=info msg="Couldn't read migration file '/connect/config/development-migration.gcfg': stat /connect/config/development-migration.gcfg: no such file or directory"
time="2022-12-12T18:08:08.679Z" level=fatal msg="Error: Unable to load configuration: Couldn't read configuration file '/connect/config/development.gcfg': warning:\ncan't store data at section \"Logging\", variable \"Foo\"\n"
start-licensed.sh: Deactivating license ...
error: Recipe `start` failed with exit code 1
```

PR #126, opened 2023-12-04
Support S3 Objects List ETag map

jonyoder:
Adds a `BucketObjectsMap` method that returns a map of S3 keys to ETag values. This method is nearly identical to the existing `BucketObjects` method, but differs slightly to return a map with the ETag values.

### DIFF
--- a/pkg/rsstorage/servers/s3server/s3_enumeration.go
+++ b/pkg/rsstorage/servers/s3server/s3_enumeration.go
@@ -19,6 +19,7 @@ import (
 type AwsOps interface {
 	BucketDirs(bucket, s3Prefix string) ([]string, error)
 	BucketObjects(bucket, s3Prefix string, concurrency int, recursive bool, reg *regexp.Regexp) ([]string, error)
+	BucketObjectsETagMap(bucket, s3Prefix string, concurrency int, recursive bool, reg *regexp.Regexp) (map[string]string, error)
 }
 
 type DefaultAwsOps struct {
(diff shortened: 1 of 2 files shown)


pkg/rsstorage/servers/s3server/s3_enumeration.go
```diff
@@ -157,6 +158,114 @@ func (a *DefaultAwsOps) BucketObjects(bucket, s3Prefix string, concurrency int,
 	}
 }
 
+func (a *DefaultAwsOps) BucketObjectsMap(bucket, s3Prefix string, concurrency int, recursive bool, reg *regexp.Regexp) (map[string]string, error) {
```
jonyoder, line 161: No need to review this whole method in detail. I'll annotate what changed, apart from the signature.

pkg/rsstorage/servers/s3server/s3_enumeration.go
```diff
@@ -157,6 +158,114 @@ func (a *DefaultAwsOps) BucketObjects(bucket, s3Prefix string, concurrency int,
...
+						wg.Done()
+						// TODO: `nm` may always be blank when there are no
+						// contents, so this conditional may be unnecessary.
+						if nm == "" {
+							wg.Done()
+						}
+						return
+					}
+
+					bm := getObjectsAllMap(resp, s3Prefix, reg)
```
jonyoder, line 231: This is the only line that differs materially from `BucketObjects`. It calls the new `getObjectsAllMap` method instead of `getObjectsAll`.

pkg/rsstorage/servers/s3server/s3_enumeration.go
```diff
@@ -176,3 +285,21 @@ func getObjectsAll(bucketObjectsList *s3.ListObjectsOutput, s3Prefix string, reg
...
+	binaryMeta := make(map[string]string)
+
+	for _, key := range bucketObjectsList.Contents {
+
+		if reg != nil {
+			if s := reg.FindStringSubmatch(*key.Key); len(s) > 1 {
+				binaryMeta[strings.TrimPrefix(s[1], s3Prefix)] = *key.ETag
+			}
+		} else {
+			binaryMeta[strings.TrimPrefix(*key.Key, s3Prefix)] = *key.ETag
```
jonyoder, line 299: This method is nearly identical to the existing `getObjectsAll` method except that it returns a map and sets the map key/value pairs here instead of simply appending to a string array.

pkg/rsstorage/servers/s3server/s3_enumeration.go
```diff
@@ -19,6 +19,7 @@ import (
 type AwsOps interface {
 	BucketDirs(bucket, s3Prefix string) ([]string, error)
 	BucketObjects(bucket, s3Prefix string, concurrency int, recursive bool, reg *regexp.Regexp) ([]string, error)
+	BucketObjectsMap(bucket, s3Prefix string, concurrency int, recursive bool, reg *regexp.Regexp) (map[string]string, error)
```
glin, line 22: Looks good, I'd just suggest making this more obvious that it's map of ETag values. That could be a comment, but maybe renaming the method works better. Or if it's possible we'd need more metadata, a map of structs that might someday contain LastModified time or whatever.

glin (approved):

PR #133, opened 2024-02-13
Create dependabot.yml

shepherdjerred:
It would be good to keep the platform-lib dependencies up-to-date since Package Manager relies on it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -103,3 +103,111 @@ updates:
         dependency-type: "development"
     reviewers:
       - "rstudio/ppm"
+  - package-ecosystem: "gomod"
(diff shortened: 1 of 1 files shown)


jonyoder (changes requested):
There are several packages missing:

```
pkg/rselection/impls/pgx
pkg/rselection/impls/local
pkg/rsstorage/servers/s3server
pkg/rsstorage/servers/postgres
pkg/rsstorage/servers/file
pkg/rsnotify/listeners/postgrespgx
pkg/rsnotify/listeners/postgrespq
pkg/rsnotify/listeners/local
pkg/rsqueue/impls/database
```

jonyoder (approved):

PR #186, opened 2024-06-17
rslogtest.LoggerMock: synchronize formatted messages

### LINKED ISSUE #185: rslogtest.LoggerMock needs synchronization
The `rslogtest.LoggerMock` type is not safe to be called from multiple goroutines. In particular, its `stringCalls` field is updated in each call and lacks synchronization.

aronatkins:
Addresses some concurrency problems with `rslogtest.LoggerMock`.

```bash
MODULE=pkg/rslog/rslogtest just test -race
```

Test code may use loggers across multiple goroutines, but generally avoids inspecting the logged information while the code-under-test is running. This work changes how logged messages are recorded, which frequently occurs within multiple goroutines.

Aspects of `mock.Mock` are not concurrency-safe; this change does not address those problems.

Fixes #185

### DIFF
--- a/pkg/rslog/rslogtest/mocks.go
+++ b/pkg/rslog/rslogtest/mocks.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"log"
 	"regexp"
+	"sync"
 
 	"github.com/rstudio/platform-lib/pkg/rslog"
 	"github.com/stretchr/testify/mock"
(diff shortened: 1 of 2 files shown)


marcosnav (approved):

aronatkins:
> Looks good to me! Do you think there would be any benefit to using an `RWMutex` instead?

A simple `sync.Mutex` is fine in this case. This is helper code to assist with tests.

The majority of references are writes. Additionally, test code really should not be reading the state of `mock.Mock` objects while there is active concurrency because of problems like https://github.com/stretchr/testify/issues/1128. 

If folks need to actively observe log messages while they're occurring, they should use some other strategy, such as a logger which communicates incoming messages over a channel.

PR #196, opened 2024-08-29
rsstorage: return original resolver err when S3 upload fails during `Put()`

glin:
When calling `StorageServer.Put()` for S3, a generic S3 upload error was always being returned when the resolver failed:

```
Something went wrong uploading to S3. You may want to check your configuration, error: ReadRequestBody: read upload data failed
caused by: Package badmissingpackage not found
```

This was making it hard to handle errors, and was also misleading because the failure occurred before even attempting to upload.

We could maybe return a wrapped error here instead? https://github.com/rstudio/platform-lib/blob/133c6151f117e37eecc4ed3edaece7e84b5a1edd/pkg/rsstorage/servers/s3server/s3_service.go#L126

But this also seems to work, returning the original resolver error if an upload failed because of the resolver.

This is for the P3M issue with invalid PyPI package index requests returning these S3 errors as unhandled 500s, rather than a 404.

### DIFF
--- a/pkg/rsstorage/servers/s3server/server.go
+++ b/pkg/rsstorage/servers/s3server/server.go
@@ -251,17 +251,17 @@ func (s *StorageServer) Put(resolve types.Resolver, dir, address string) (string
 	// Resolve (read) the item using the piped writer
 	var wdir string
 	var waddress string
+	var resolverErr error
 	go func() {
 		// Make sure an EOF gets sent to the pipe writer. Without this, we end
 		// up hanging forever while the uploader reads from the pipe.
-		var err error
-		wdir, waddress, err = resolve(w)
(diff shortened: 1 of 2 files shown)


jonyoder (approved):

jonyoder:
Looks good. In case you missed it, we recently changed this project to be a single Go module (with the exception of `rslog`, which is still separate). When you merge this, you'll just need to push one new tag:

```bash
git tag v2.0.2
git push origin --tags
```

glin:
@jonyoder Oh gotcha, I was going to ask about that. Do we still do releases? Or tags only?

jonyoder:
I haven't kept up with releases, but you're welcome to create one if you want.

PR #198, opened 2024-09-09
Fix: Typecast blocksize to work on both Unix and Linux

tylfin:


nunyunuymi (approved):

pkg/rsstorage/servers/file/server_unix.go
```diff
@@ -15,7 +15,7 @@ func Statfs(path string) (*StatfsData, error) {
 		return nil, err
 	}
 	return &StatfsData{
-		Bsize:  fs.Bsize,
```
jmwoliver, line 18: Could we have instead just changed the type of Bsize to uint32 instead of typecasting?
https://github.com/rstudio/platform-lib/blob/b94d7fd7393c5e690783152277c54cba0d191579/pkg/rsstorage/servers/file/server.go#L122

PR #202, opened 2024-10-21
feat: option to initialize storage servers without chunker

jmwoliver:
## Description

The goal with this change is to allow storage servers to be initialized without needing a database to be running first.

By allowing the storage server to start without the chunker, it can still use `Put()` without chunking. When `WriteChunked()` or `ReadChunked()` are needed by the storage server, it first checks that the chunker is not nil. If it is, it will throw an error.

Initial work needed for https://github.com/rstudio/package-manager/issues/13651.


tylfin:
What does the Chunker use the database for? Is there an opportunity to decouple the two entirely?

PR #209, opened 2025-01-24
debug logger: remove logger callbacks

aronatkins:
All debug logger instances were previously attached to a global set of callbacks. This was done so loggers could be informed if they were ever enabled dynamically. Unfortunately, loggers were never removed from this set of callbacks, which meant that those loggers could never be reclaimed by the garbage collector. Debug loggers are often constructed as the program runs, adding fields and values to assist with tracing. This meant that programs often did not have a static set of loggers, but were always creating new debug loggers, which were always seen as live objects.

Now, debug loggers always ask their enabled status from the map tracking enabled state. This incurs the cost of a mutex-read lock and map lookup at runtime.

This change is associated with Connect issue 29616

### DIFF
--- a/pkg/rslog/debug.go
+++ b/pkg/rslog/debug.go
@@ -1,65 +1,95 @@
 package rslog
 
-// Copyright (C) 2022 by RStudio, PBC.
+// Copyright (C) 2025 by Posit Software, PBC.
 
 import (
 	"sync"
 )
 
+// ProductRegion is a numerical value assigned to an area of the product.
 type ProductRegion int
 
-// regionNames translates the enum region to its string equivalent. This is
-// used for display (when logging) and also when processing the configuration
-// values.
-var regionNames map[ProductRegion]string
-
-type callbacksArr []func(flag bool)
-
-var regionCallbacks = map[ProductRegion]callbacksArr{}
-var regionsEnabled = map[ProductRegion]bool{}
+// debugMutex protects access to product debug region/enabled information.
 var debugMutex sync.RWMutex
 
+// debugRegions relates a numerical product debug region to its string
+// equivalent. Used for display (when logging) and when processing
+// configuration values.
+//
+// Protected by debugMutex.
+var debugRegions map[ProductRegion]string = map[ProductRegion]string{}
+
+// debugEnabled records product debug regions which enable debug logging.
+//
+// Protected by debugMutex.
+var debugEnabled map[ProductRegion]bool = map[ProductRegion]bool{}
+
+// RegisterRegions registers product debug regions. Called once at program
+// startup.
 func RegisterRegions(regions map[ProductRegion]string) {
-	regionNames = regions
+	debugMutex.Lock()
+	defer debugMutex.Unlock()
+
+	debugRegions = regions
 }
 
+// RegionNames returns the names of all registered product debug regions.
 func RegionNames() []string {
+	debugMutex.RLock()
+	defer debugMutex.RUnlock()
+
 	var names []string
-	for _, name := range regionNames {
+	for _, name := range debugRegions {
 		names = append(names, name)
 	}
 	return names
 }
 
+// Regions returns the numerical product debug regions.
 func Regions() []ProductRegion {
+	debugMutex.RLock()
+	defer debugMutex.RUnlock()
+
 	var regions []ProductRegion
-	for r := range regionNames {
+	for r := range debugRegions {
 		regions = append(regions, r)
 	}
 	return regions
 }
 
-func RegionByName(text string) ProductRegion {
-	for region, name := range regionNames {
-		if name == text {
+// RegionByName returns the numerical product debug region associated with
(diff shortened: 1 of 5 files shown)


pkg/rslog/debug.go
```diff
@@ -6,106 +6,125 @@ import (
...
 
+// ProductRegion is a numerical value assigned to an area of the product.
 type ProductRegion int
 
-// regionNames translates the enum region to its string equivalent. This is
-// used for display (when logging) and also when processing the configuration
-// values.
-var regionNames map[ProductRegion]string
+// debugMutex protects access to product debug region/enabled information.
+var debugMutex sync.RWMutex
```
aronatkins, line 13: Renamed the globals to clearly indicate that they are only for debug log handling.

pkg/rslog/debug.go
```diff
@@ -6,106 +6,125 @@ import (
 	"sync"
 )
 
+// ProductRegion is a numerical value assigned to an area of the product.
```
aronatkins, line 9: Added docs throughout the file.

pkg/rslog/debug.go
```diff
@@ -6,106 +6,125 @@ import (
...
+// debugEnabled records product debug regions which enable debug logging.
+//
+// Protected by debugMutex.
+var debugEnabled map[ProductRegion]bool = map[ProductRegion]bool{}
 
+// RegisterRegions registers product debug regions. Called once at program
+// startup.
 func RegisterRegions(regions map[ProductRegion]string) {
-	regionNames = regions
+	debugMutex.Lock()
```
aronatkins, line 30: Used the mutex for ALL `debugRegions` and `debugEnabled` accesses. We no longer assume that some functions are only called at program startup. Consistent use of the mutex is easier to explain that implicit conventions.

pkg/rslog/debug.go
```diff
@@ -6,106 +6,125 @@ import (
...
 
 	// match debug log region to list
 	for _, region := range regions {
 		if region == 0 {
 			continue
 		}
-		regionsEnabled[region] = true
-		regionName := RegionName(region)
+		debugEnabled[region] = true
+		regionName := debugRegions[region]
```
aronatkins, line 101: `InitDebugLogs()` directly accesses `debugRegions` because `RegionName()` now uses `debugMutex`; we need `InitDebugLogs()` to hold the mutex until its work is complete.

pkg/rslog/debug.go
```diff
@@ -1,111 +1,130 @@
...
-func RegionByName(text string) ProductRegion {
-	for region, name := range regionNames {
-		if name == text {
+// RegionByName returns the numerical product debug region associated with
+// name. Returns zero when there is no match.
+func RegionByName(name string) ProductRegion {
+	debugMutex.RLock()
+	defer debugMutex.RUnlock()
+
+	for region, region_name := range debugRegions {
```
dbkegley, line 66: camel case?

pkg/rslog/debug_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+import (
+	"slices"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+// Copyright (C) 2025 by Posit Software, PBC.
```
dbkegley, line 13:
```suggestion
```

pkg/rslog/debug.go
```diff
@@ -118,86 +137,73 @@ type DebugLogger interface {
...
 
-// WithSubRegion returns a debug logger with further specificity
-// via sub_region key:value. E.g "region": "LDAP", "sub_region": "membership scanner"
+// WithSubRegion returns a debug logger having an additional "sub_region"
+// field. The returned logger is associated with the same product debug
+// region.
+//
+// Equivalent to `debugLogger.WithField("sub_region", subregion)`
 func (l *debugLogger) WithSubRegion(subregion string) DebugLogger {
 	newLgr := l.Logger.WithField("sub_region", subregion)
```
dbkegley, line 203: This isn't really relevant for this PR but I don't understand why `WithSubRegion` exists. Why not just call `WithFields()` like we do in the tests down below?

```go
                gr.WithFields(Fields{
			"region":     "alfa",
			"sub_region": "drinks",
		})
```

dbkegley (approved):

pkg/rslog/debug.go
aronatkins: I do not want to alter the public API.

pkg/rslog/debug_test.go
aronatkins: Ugh. It was in the wrong place, failed the linter, then I duplicated it. So awesome.

pkg/rslog/debug.go
aronatkins: The API exists because folks wanted a well-known field name.

PR #224, opened 2025-06-24
Migrate to Go AWS SDK V2

CDRayn:
## About

This PR migrates `platform-lib` to the v2 version of the Go AWS SDK. This migrations is necessary due to the v1 version of the SDK being deprecated and AWS dropping support for it on 7/31/2025. The high level changes made as part of the migration will be summarized below, but the [official migration guide](https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/migrate-gosdk.html) is a good resource for reviewers who wish to better understand the context and motivation behind the changes in this PR.

### Configuration Loading & Initialization

The v2 version of the SDK replaces the `session` package and its associated functionality with the simplified [config](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/config). The [Configuration Loading](https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/migrate-gosdk.html#configuration-loading) section of the migration guide explains the changes to this area between versions of the SDK. For flexibility, this PR updates `platform-lib` to leave parsing / consuming the AWS config to the consumer of this library and expects an `Options` interface passed to it.

### S3 Copier

The S3 copier implementation was removed by this PR since the v2 version of the SDK adds builtin support for moving and copying objects between S3 buckets.

### Go context parameter

v2 of the SDK adds support for the use of Go's `ctx context.Context` interface. This PR adds `ctx context.Context` to the signatures of functions that may result in network operations or long running tasks, and which may need to be cancelled.

### Formatting

For readability and clarity, lines in the code base that exceeded 120 characters in length were wrapped across multiple lines. Some variable and symbol declarations were updated because their name collided with package imports or Go builtin items.


pkg/rscache/file.go
```diff
@@ -185,12 +194,18 @@ func (o *fileCache) Head(ctx context.Context, resolver ResolverSpec) (size int64
 
 		// Wait
 		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
```
nodivbyzero, line 200: do we need `default` here?

CDRayn: @nodivbyzero Let me experiment with it some more, but if I'm remembering correctly, I was experiencing some test failures if I left out the `default` statement.

CDRayn: @nodivbyzero it looks like removing the `default` statement causes the following code to be unreachable and results in a deadlock.

pkg/rscache/memory_backed_file_cache.go
```diff
@@ -14,6 +14,7 @@ import (
 )
 
 func decompressAndDecodeGob(reader io.ReadCloser, compressed bool, typeExample interface{}) (result interface{}, err error) {
+	// TODO: Handle this error
```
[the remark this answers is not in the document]
nodivbyzero: Looks great!

glin:
I haven't reviewed yet, but just from a quick look, there are a lot of files changed only for the copyright update, so it's hard to even browse the changes in the diff. Chunking up/squashing the commits so the copyright update only updates are separated would probably help a lot. Same with the other changes, so it's easier to review a single commit without the copyrights.

jonyoder (commented):
One preliminary question - have we tried these changes with PPM? You can tweak the PPM `go.mod` to include the `platform-lib` modules from your local project directory and do any PPM refactoring necessary, then see if you run into issues. That may help catch any blockers early instead of later.

CDRayn:
> One preliminary question - have we tried these changes with PPM? You can tweak the PPM `go.mod` to include the `platform-lib` modules from your local project directory and do any PPM refactoring necessary, then see if you run into issues. That may help catch any blockers early instead of later.

@jonyoder I haven't yet but that is my plan once all the integrations tests are working. Once all the tests pass, I'll use PPM as a final integration / end-to-end test before these changes get merged.

jstruzik:
@CDRayn I know SDK updates can be tough, especially as the repo is constantly changing, but is there a way to break this PR down a bit further like Greg is suggesting? I imagine a separate copyright work would be approved and merged rather quickly

nunyunuymi (approved):

pkg/rsstorage/internal/integration_test/integration_test.go
```diff
@@ -126,20 +139,25 @@ func (s *StorageIntegrationSuite) TearDownTest(c *check.C) {
...
+	s3Svc, err := s3server.NewS3Wrapper(
+		s3.Options{
+			Region:             "us-east-1",
+			EndpointOptions:    s3.EndpointResolverOptions{DisableHTTPS: true},
+			EndpointResolverV2: &Resolver{URL: endpointURL},
+			UsePathStyle:       true,
+			Credentials:        credentials.NewStaticCredentialsProvider("minio", "miniokey", ""),
+			//ClientLogMode:      aws.LogRequestWithBody | aws.LogResponseWithBody,
+			//RetryMaxAttempts:   1,
+			//Logger: logging.NewStandardLogger(os.Stdout),
```
nodivbyzero: debug code?

CDRayn: @nodivbyzero, ah I added those for debugging purposes but decided to leave them in for future use. I added a comment in ea7e1bb to explain that they should only be enabled for debugging since they are a lot of noise to the tests' output.

pkg/rscache/memory_backed_file_cache.go
```diff
@@ -24,7 +23,6 @@ func decompressAndDecodeGob(reader io.ReadCloser, compressed bool, typeExample i
 		if err != nil {
 			return
 		}
-		defer uncompressedReader.Close()
```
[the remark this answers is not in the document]
CDRayn: @nodivbyzero I originally removed that because it looked like it was causing a duplicate call to `Close()` but after checking again, that looks like it's no longer the case. I added it back in e13033c and wrapped the deferred call to `Close()` in a closure to properly handle the error.

pkg/rscache/memory_backed_file_cache.go
```diff
@@ -7,15 +7,14 @@ import (
...
 	"encoding/gob"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
 	"time"
 )
 
 func decompressAndDecodeGob(reader io.ReadCloser, compressed bool, typeExample interface{}) (result interface{}, err error) {
-	defer reader.Close()
```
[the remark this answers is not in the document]
CDRayn: @nodivbyzero After reviewing it again, in 3ab3346 I added a single call to `Close()` for just the underlying `ReadCloser`. Calling `Close()` just on `ReadCloser` fixed the error caused by calling `Close()` twice.

PR #251, opened 2025-12-30
Respect context in file cache `Get()` and `Head()`

glin:
For https://github.com/rstudio/package-manager/issues/16862

Fix an issue with the file cache's `Get()` and `Head()` not respecting context, e.g. not properly timing out given a context with timeout.


CDRayn (approved):
Looks good to me. Approved.

PR #256, opened 2026-02-06
Finish migrating platform-lib to slog

glin:
platform-lib is currently partially migrated to slog, with a lot of `log.Printf()`'s remaining.

This finishes up the migration for all packages within it, because error logs getting logged to `INFO` has been causing problems:
```sh
INFO Received error on queue addressed work notification channel: ... connection timed out

INFO Error pushing leader assumption work to queue: Duplicate address 

INFO Error verifying cluster integrity: node list length differs. Store node count 2 does not match leader count 1
```

**Note:** this is just a simple find-and-replace style swap of `log.Printf()` to `slog.Info(fmt.Sprintf(...))`. It's not idiomatic slog using key-val pairs but I wanted to keep this as simple to review as possible. Most of the other slog uses in platform-lib use key-val pairs either. I think we just migrate those either all at once or when updating individual files in the future.

### DIFF
--- a/pkg/rselection/impls/pgx/leader.go
+++ b/pkg/rselection/impls/pgx/leader.go
@@ -221,7 +221,7 @@ func (p *PgxLeader) verify(vCh chan bool) {
 			// e.g. when restarting a cluster node and the node list is temporarily out of sync:
 			// Error verifying cluster integrity: node list length differs. Store node count 3 does not match leader count 2
 			// It would be great to be able to distinguish between expected/unexpected errors here and log accordingly.
-			slog.Error(fmt.Sprintf("Error verifying cluster integrity: %s", *err))
+			slog.Error("Error verifying cluster integrity", "error", *err)
 			vCh <- false
 		} else {
 			vCh <- true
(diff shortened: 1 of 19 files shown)


pkg/rselection/impls/pgx/leader.go
```diff
@@ -218,7 +217,11 @@ func (p *PgxLeader) verify(vCh chan bool) {
 	// Upon exit, notify channel with `true` or `false` depending upon error status
 	defer func(err *error) {
 		if *err != nil {
-			log.Printf("Error verifying cluster integrity: %s", *err)
+			// TODO this should be worded better because sometimes this error is normal and expected,
+			// e.g. when restarting a cluster node and the node list is temporarily out of sync:
+			// Error verifying cluster integrity: node list length differs. Store node count 3 does not match leader count 2
+			// It would be great to be able to distinguish between expected/unexpected errors here and log accordingly.
+			slog.Error(fmt.Sprintf("Error verifying cluster integrity: %s", *err))
```
glin: Just calling this one out as it may introduce ERROR log noise in logs when upgrading/restart cluster nodes. But I figured it's much more important to have these surfaced as ERRORs when it does indicate an actual error.

nodivbyzero (approved):

PR #261, opened 2026-03-05
fix: fail fast when task is registered after Handle()

jonyoder:
## Summary

- Add a runtime guard in `GenericTaskHandler.Register()` that fatals if a task is registered after `Handle()` has already been called
- `Handle()` iterates the task map once to start goroutines — any task registered afterward exists in the map but never gets a goroutine, silently failing
- This was the root cause of rstudio/package-manager#17126, where the OpenVSX auto-sync task was registered ~270 lines after `Handle()` had run

## Changes

- `Register()` now checks if `h.cancel != nil` (set by `Handle()`) and fatals with a clear message if so
- Existing `log.Fatalf` calls use a package-level `logFatalf` variable to allow test overrides
- Added `TestRegisterAfterHandleFatals` test

## Test plan

- [x] Existing `TestGenericTaskHandler` passes (registration before `Handle()` still works)
- [x] New `TestRegisterAfterHandleFatals` verifies the guard triggers

Related: rstudio/package-manager#17328

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/rselection/taskhandler.go
+++ b/pkg/rselection/taskhandler.go
@@ -67,6 +67,9 @@ func NewGenericTaskHandler(cfg GenericTaskHandlerConfig) *GenericTaskHandler {
 	}
 }
 
+// Register adds a task to the handler. It must be called before Handle().
+// Calling Register after Handle() or registering the same task name twice
+// will cause the program to exit via log.Fatalf.
 func (h *GenericTaskHandler) Register(name string, task Task) {
 	if h.cancel != nil {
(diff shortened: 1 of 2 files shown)


CDRayn (approved):
I tend to dislike the idea of libraries / packages causing a panic or making calls to `os.Exit()` via `Fatal*()` but I don't think there is an alternative here without having to make a bunch of other changes / refactors. So I think this instance of it is fine. We should make sure we document the behavior of `Register()` now that it can potentially cause an application to exit.

PR #276, opened 2026-05-12
Update S3 based storage to accept AWS S3 client interface for instantiation

CDRayn:
## About

This PR refactors the S3 backed storage server in `pkg/rsstorage/servers/s3server` to accept a AWS S3 client interface for instantiation. This is part of a bugfix for https://github.com/rstudio/package-manager/issues/17944 that resulted in `S3 PutObject` operations to fail if the S3 bucket used for the storage server had Object Lock enabled. This was a result of upgrading from the deprecated AWS Go SDK to the latest AWS Go SDK, which handles constructing configurations and included default options differently than the deprecated SDK. This resulted in the `RequestChecksumCalculation` being unset, when it needed to be set to `RequestChecksumCalculationWhenSupported` when Object Lock is enabled for the underlying PR bucket.

## Fixes

https://github.com/rstudio/package-manager/issues/17944

## Summary of changes

- Added the `S3API` interface as a subset of only the required methods of the AWS S3 client interface.
- Updated `NewEncryptedS3Wrapper()` to accept the `S3API` interface instead of S3 Options
- Updated `NewS3Wrapper()` to accept an `S3API` interface that is intended to be satisfied by the S3 client interface
- Pivoted `platform-lib/rsstorage` to accept a S3 client for instantiation and leave S3 configuration to consumers of the package
- Updated functions in `pkg/rsstorage/servers/s3server` to accept the new interfaces but return concrete types as to avoid confusion

### DIFF
--- a/pkg/rsstorage/servers/s3server/s3_encrypted_service.go
+++ b/pkg/rsstorage/servers/s3server/s3_encrypted_service.go
@@ -6,63 +6,47 @@ import (
 	"context"
 	"fmt"
 
-	encryptClient "github.com/aws/amazon-s3-encryption-client-go/v3/client"
-	"github.com/aws/amazon-s3-encryption-client-go/v3/materials"
-	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/feature/s3/manager"
-	"github.com/aws/aws-sdk-go-v2/service/kms"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/rstudio/platform-lib/v3/pkg/rsstorage/internal"
 )
 
-type encryptedS3Service struct {
-	client *encryptClient.S3EncryptionClientV3
+type EncryptedS3Service struct {
+	client S3API
 }
 
-func NewEncryptedS3Wrapper(s3Opts s3.Options, kmsOpts kms.Options, keyID string) (S3Wrapper, error) {
-	if s3Opts.Region == "" || kmsOpts.Region == "" {
-		return nil, fmt.Errorf("AWS configuration option 'region' is required")
+// NewEncryptedS3Wrapper constructs a S3Wrapper backed by a client-side
+// encryption-aware S3 client (typically *encryptClient.S3EncryptionClientV3).
+// Callers are responsible for constructing the encryption client themselves
+// — including the underlying *s3.Client, *kms.Client, and
+// CryptographicMaterialsManager — and passing it in here.
+func NewEncryptedS3Wrapper(client S3API) (*EncryptedS3Service, error) {
+	if client == nil {
+		return nil, fmt.Errorf("an S3 client must be provided to NewEncryptedS3Wrapper")
 	}
-
-	s3Client := s3.New(s3Opts)
-	kmsClient := kms.New(kmsOpts)
-
-	cmm, err := materials.NewCryptographicMaterialsManager(materials.NewKmsKeyring(kmsClient, keyID))
-	if err != nil {
-		return nil, fmt.Errorf("error encounted while initializing crytographic manager: %w", err)
-	}
-	client, err := encryptClient.New(s3Client, cmm)
-	if err != nil {
-		return nil, fmt.Errorf("unable to initialize S3 encryption client: %w", err)
-	}
-
-	return &encryptedS3Service{
+	return &EncryptedS3Service{
 		client: client,
 	}, nil
 }
 
-func (s *encryptedS3Service) getConfig() config.Config {
-	return s.client.Options
-}
-
-func (s *encryptedS3Service) CreateBucket(ctx context.Context, input *s3.CreateBucketInput) (*s3.CreateBucketOutput, error) {
+func (s *EncryptedS3Service) CreateBucket(ctx context.Context, input *s3.CreateBucketInput) (*s3.CreateBucketOutput, error) {
 	return s.client.CreateBucket(ctx, input)
 }
 
-func (s *encryptedS3Service) DeleteBucket(ctx context.Context, input *s3.DeleteBucketInput) (*s3.DeleteBucketOutput, error) {
+func (s *EncryptedS3Service) DeleteBucket(ctx context.Context, input *s3.DeleteBucketInput) (*s3.DeleteBucketOutput, error) {
 	return s.client.DeleteBucket(ctx, input)
 }
 
-func (s *encryptedS3Service) HeadObject(ctx context.Context, input *s3.HeadObjectInput) (*s3.HeadObjectOutput, error) {
+func (s *EncryptedS3Service) HeadObject(ctx context.Context, input *s3.HeadObjectInput) (*s3.HeadObjectOutput, error) {
 	return s.client.HeadObject(ctx, input)
 }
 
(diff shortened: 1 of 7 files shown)


CDRayn:
One thing I want to point out and get feedback on, is the refactor of `pkg/rsstorage/servers/s3server` to accept a AWS S3 client interface for instantiation as part of this bugfix means this is a breaking change and will require a new major semantic version. I think this is justified but I want to open it up to others on the team.

pkg/rsstorage/servers/s3server/s3_service.go
```diff
@@ -30,33 +57,35 @@ type S3Wrapper interface {
...
-func NewS3Wrapper(s3Opts s3.Options) (S3Wrapper, error) {
-	if s3Opts.Region == "" {
-		return nil, fmt.Errorf("'region' field of ConfigS3 is required")
+// NewS3Wrapper constructs a S3Wrapper backed by the provided S3 client. The
+// caller is responsible for configuring and constructing the client (e.g.
+// region, credentials, endpoint). Passing in the client rather than
+// configuration values lets callers wrap or replace the AWS SDK client with
+// their own implementation, which is useful for testing and for layering on
+// behaviors such as Object Lock-aware deletes.
+func NewS3Wrapper(client S3API) *DefaultS3Wrapper {
```
nodivbyzero: This is a breaking change (return type changed from interface to concrete type) for any downstream code that relied on the interface return type.

pkg/rsstorage/servers/s3server/s3_encrypted_service.go
```diff
@@ -6,63 +6,47 @@ import (
...
-func NewEncryptedS3Wrapper(s3Opts s3.Options, kmsOpts kms.Options, keyID string) (S3Wrapper, error) {
-	if s3Opts.Region == "" || kmsOpts.Region == "" {
-		return nil, fmt.Errorf("AWS configuration option 'region' is required")
+// NewEncryptedS3Wrapper constructs a S3Wrapper backed by a client-side
+// encryption-aware S3 client (typically *encryptClient.S3EncryptionClientV3).
+// Callers are responsible for constructing the encryption client themselves
+// — including the underlying *s3.Client, *kms.Client, and
+// CryptographicMaterialsManager — and passing it in here.
+func NewEncryptedS3Wrapper(client S3API) (*EncryptedS3Service, error) {
+	if client == nil {
```
nodivbyzero, line 26: `NewEncryptedS3Wrapper` has a nil client check:
```go
if client == nil {
    return nil, fmt.Errorf("an S3 client must be provided to NewEncryptedS3Wrapper")
}
```

But `NewS3Wrapper` does not:
```go
func NewS3Wrapper(client S3API) *DefaultS3Wrapper {
    return &DefaultS3Wrapper{
        client: client,
    }
}
```

If `nil` is passed to `NewS3Wrapper`, it will panic on first use rather than failing fast at construction.

**Recommendation**: Add a nil check to `NewS3Wrapper` for consistency and better error messages.

pkg/rsstorage/servers/s3server/s3_encrypted_service.go
```diff
@@ -6,63 +6,47 @@ import (
...
 	"github.com/aws/aws-sdk-go-v2/feature/s3/manager"
-	"github.com/aws/aws-sdk-go-v2/service/kms"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/rstudio/platform-lib/v3/pkg/rsstorage/internal"
 )
 
-type encryptedS3Service struct {
-	client *encryptClient.S3EncryptionClientV3
+type EncryptedS3Service struct {
```
nodivbyzero: Type naming inconsistency
- `defaultS3Wrapper` → `DefaultS3Wrapper` (exported)
- `encryptedS3Service` → `EncryptedS3Service` (exported)

The second type is now `Service` while the first is `Wrapper`. Consider naming them consistently:
- `DefaultS3Wrapper` and `EncryptedS3Wrapper`, or
- `DefaultS3Service` and `EncryptedS3Service`

pkg/rsstorage/servers/s3server/s3_service.go
```diff
@@ -191,9 +220,11 @@ func (s *defaultS3Wrapper) MoveObject(ctx context.Context, oldBucket, oldKey, ne
 	return out, nil
 }
 
-func (s *defaultS3Wrapper) ListObjects(ctx context.Context, input *s3.ListObjectsV2Input) (*s3.ListObjectsV2Output, error) {
+func (s *DefaultS3Wrapper) ListObjects(ctx context.Context, input *s3.ListObjectsV2Input) (*s3.ListObjectsV2Output, error) {
 	// In AWS SDK v2, we need to handle pagination manually
-	// Create a paginator to iterate through all pages
+	// Create a paginator to iterate through all pages. S3API satisfies the
+	// s3.ListObjectsV2APIClient interface implicitly because it declares
```
nodivbyzero, line 229: `ListObjects` pagination behavior differs between implementations

The `DefaultS3Wrapper.ListObjects` collects all pages:
```go
paginator := s3.NewListObjectsV2Paginator(s.client, input)
// ... iterates all pages
```

But `EncryptedS3Service.ListObjects` just calls through:
```go
return s.client.ListObjectsV2(ctx, input)
```

This behavioral difference existed before but is worth noting - the encrypted version won't paginate automatically while the default version does.

pkg/rsstorage/servers/s3server/s3_encrypted_service.go
CDRayn: Updated in 78a37d8.

pkg/rsstorage/servers/s3server/s3_encrypted_service.go
CDRayn: Updated in 78a37d8.

pkg/rsstorage/servers/s3server/s3_service.go
CDRayn: I think it's worth updating as part of this PR. Updated in 78a37d8.

pkg/rsstorage/servers/s3server/s3_service.go
CDRayn: Changing the return type of this function from a interface to a type technically isn't a breaking change since the type it was changed to implements the interface that was originally breaks. This practice of [accepting interfaces and returning types](https://duncanleung.com/go-idiom-accept-interfaces-return-types/) is considered an idiomatic Go practice. That was the motivation for changing the return type.

The change to the function argument is certainly a breaking change like you pointed out. I'll make sure this gets documented as a breaking change and highlighted in any change log.

CDRayn: As for breaking changes impacting consumers of this package, I searched our codes bases and the only other active project besides Package Manager that uses `rstudio/platform-lib/pkg/rsstorage` is [r-manifest](https://github.com/rstudio/r-manifest). This is the [rstudio org search](https://github.com/search?q=org%3Arstudio+rsstorage+language%3AGo+NOT+repo%3Arstudio%2Fpackage-manager+NOT+repo%3Arstudio%2Fplatform-lib&type=code&l=Go&p=1) and [posit-dev org](https://github.com/search?q=org%3Aposit-dev+rsstorage&type=code) search I used to confirm this.

nodivbyzero (approved):

PR #280, opened 2026-06-02
Recover cluster leader from Postgres failover

jonyoder:
## Problem

When the Postgres database behind a multi-node PPM cluster fails over (e.g. an AWS RDS maintenance window), `PgxLeader` can wedge: it keeps emitting pings and considers itself leader, but stops running scheduled work. Two collaborating bugs cause this:

1. `Lead()` never returns on persistent failure — its only exit was the stop channel, so the supervisor's leader→follower recovery path was never taken.
2. The in-memory node map drifts from the store during the outage and `verify()`'s strict length check then fails permanently, silently skipping every scheduled task.

This is the underlying cause of repeated production incidents (`node list length differs ...`) downstream in Posit Package Manager (rstudio/package-manager#17889).

## Changes

- **`clusterIntegrityErr()`** — a shared cluster-integrity check that reconciles the in-memory node map *asymmetrically*: prune nodes that have left the store (safe), but never add store-only nodes. Reachable nodes are re-added by `handlePingResponse` when they answer a ping, which preserves the split-brain guard. `verify()` becomes a thin wrapper over it.
- **`evaluateHealth()` + time-based step-down** — evaluated on the existing sweep tick. After `StepDownTimeout` of continuous integrity failure the leader returns from `lead()`, so `Lead()` returns nil and the consumer's supervisor re-elects. `StepDownTimeout == 0` disables step-down (legacy behavior).
- **Leveled logging** — transient mismatches (rolling restarts) log at debug; sustained failure logs at error; the disabled mode stays at debug to avoid per-tick spam.

New config field: `PgxLeaderConfig.StepDownTimeout`.

## Tests

New Postgres-free `RecoverySuite` covers the asymmetric reconcile (prune vs. error-without-adding), the step-down threshold boundary, reset-on-recovery, and the disabled (`0`) case. The existing `TestLeaderLeadInternal` integrity block was updated to the new semantics. Full Postgres-backed suite runs in CI.

### DIFF
--- a/pkg/rselection/impls/pgx/leader.go
+++ b/pkg/rselection/impls/pgx/leader.go
@@ -53,6 +53,19 @@ type PgxLeader struct {
 	pingSuccess     bool
 	pingSuccessLock sync.RWMutex
 
+	// stepDownTimeout is the duration of continuous cluster-integrity failure
+	// after which the leader steps down so a fresh election can occur. Zero
+	// disables step-down (legacy behavior).
+	stepDownTimeout time.Duration
+
+	// unhealthySince is the time the leader most recently entered an unhealthy
+	// state, or the zero value when currently healthy. Only accessed from the
+	// lead() goroutine.
+	unhealthySince time.Time
+
+	// now returns the current time; overridable in tests. Use timeNow().
+	now func() time.Time
+
 	// Stores information about active nodes and ping times
 	nodes map[string]*electiontypes.ClusterNode
 	mutex sync.RWMutex
@@ -60,6 +73,7 @@ type PgxLeader struct {
 	// Used for testing
 	loopAwareChTEST    chan bool
 	pingResponseChTEST chan bool
+	setClockTEST       func(time.Time)
 }
 
 type PgxLeaderConfig struct {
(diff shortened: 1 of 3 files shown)


pkg/rselection/impls/pgx/leader.go
```diff
@@ -201,58 +222,96 @@ func (p *PgxLeader) info() string {
...
+		storeKeys[node.Key()] = struct{}{}
+	}
+
+	// Prune in-memory nodes that are no longer present in the store.
+	// Unlike sweepNodes, we do not exempt the leader's own key here: the store
+	// includes the leader, and even if a transient store read omitted it the
+	// leader re-adds itself via its self-ping, so pruning it is self-healing.
+	for key := range p.nodes {
+		if _, ok := storeKeys[key]; !ok {
+			slog.Debug("Leader pruning node no longer present in store", "node", key)
```
CDRayn: I could go either way on this, but it might be a good idea to log this with the `INFO` level instead of `DEBUG`. In the event we need to troubleshoot a customer escalation, this information wouldn't be captured by default and would require the customer to change PPM's log level and then restart the application. Logging with the `INFO` level would help eliminate that step. The one downside is potentially increasing the noise in PPM's log. But I'm not sure how often this leader election would effective happen though.

I'll leave the decision up to you. I just wanted to bring this to your attention.

pkg/rselection/impls/pgx/leader.go
```diff
@@ -201,58 +222,96 @@ func (p *PgxLeader) info() string {
...
+		p.unhealthySince = p.timeNow()
+	}
+	dur := p.timeNow().Sub(p.unhealthySince)
+
+	// Treat brief mismatches (e.g., a rolling restart) as routine; escalate only
+	// once the condition is sustained. When step-down is disabled, stay at debug
+	// to avoid per-tick error spam since no recovery action will be taken.
+	switch {
+	case p.stepDownTimeout == 0:
+		slog.Debug("Cluster integrity check failing", "error", err, "unhealthyFor", dur)
```
CDRayn, line 302: Same comment as above about logging level. While having noisy logs can be annoying. It might be worth the tradeoff if it means customers don't have to change log levels and restart PPM in order to investigate problems. Again, I'll leave this decision up to you but wanted to still point it out.

pkg/rselection/impls/pgx/leader.go
jonyoder: Good call — agreed. A node being pruned from the leader's in-memory view is a genuine membership-change event (decommission, rolling restart, post-failover), and it only logs on an actual delete, not on every sweep tick, so the volume stays low. Bumped to `INFO` in 733396b. Left the per-tick integrity-check lines as-is (debug for transient, error for sustained) since those can be high-frequency during an outage and the error path already covers alerting.

pkg/rselection/impls/pgx/leader.go
jonyoder: Agreed on the goal (default-log visibility), with one tweak on the how. These integrity-check lines fire on every sweep tick (~5s) while unhealthy, so a blanket bump to INFO would log on every routine rolling restart and, in disabled-step-down mode, every 5s indefinitely — re-introducing the noise the leveled logging was meant to avoid.

So in 30bad6d I made it transition-based instead: the onset of an unhealthy episode logs once at INFO (`Cluster integrity check failing`), recovery logs once at INFO (`Cluster integrity recovered`), sustained failure past the grace window escalates to ERROR, and only the per-tick chatter stays at DEBUG. That captures "integrity degraded / recovered" in default logs for an escalation without per-tick spam, and it's safe in disabled mode. Let me know if you'd prefer a simpler straight INFO bump.

pkg/rselection/impls/pgx/leader_recovery_test.go
```diff
@@ -0,0 +1,143 @@
+package pgxelection
+
+// Copyright (C) 2026 by Posit Software, PBC
+
+import (
+	"errors"
+	"time"
+
+	"gopkg.in/check.v1"
```
CDRayn, line 9: This doesn't need to be changed or fixed here, but it looks like `rstudio/platform-lib` still uses `gopkg.in/check.v1` in its tests, while PPM has mostly migrated away from that package. I'll add an issue to track migrating off `gopkg.in/check.v1` and we can tackle when we have capacity for it.

CDRayn (approved):
Looks good to me. Approved.

PR #281, opened 2026-06-03
Add cluster-health observability callbacks to leader and task handler

jonyoder:
## What's Changed

Adds three optional, nil-able push callbacks so a consumer (Posit Package Manager) can emit cluster-health metrics without polling leader internals. All are no-ops when unset, so there is no behavior change for existing callers.

- **`PgxLeaderConfig.OnIntegrityResult func(error)`** — fired after every cluster integrity evaluation (`nil` = healthy).
- **`PgxLeaderConfig.OnPingResult func(success bool, t time.Time)`** — fired after every leader ping attempt. `pingNodes` was refactored into `pingNodes` + `sendPings` so the callback fires once with the final outcome, outside `pingSuccessLock`.
- **`GenericTaskHandlerConfig.OnTaskRun func(name string)`** — fired when a scheduled task runs, only after the verify gate passes (i.e. genuine leader-gated execution).

Callbacks run on the leader/task-handler goroutine and are documented as "must not block or panic."

Note: `pingNodes` now sets `pingSuccess=false` on a (practically impossible) JSON marshal failure, where previously it left the prior value; this is more correct and has no real-world impact.

## Tests

Postgres-free unit tests for each callback (integrity healthy/unhealthy, ping success/failure, task-run on a verify-gated scheduled run). Supports the cluster-observability work in rstudio/package-manager#17889.

### DIFF
--- a/pkg/rselection/impls/pgx/leader.go
+++ b/pkg/rselection/impls/pgx/leader.go
@@ -63,6 +63,9 @@ type PgxLeader struct {
 	// lead() goroutine.
 	unhealthySince time.Time
 
+	onIntegrityResult func(err error)
+	onPingResult      func(success bool, t time.Time)
+
 	// now returns the current time; overridable in tests. Use timeNow().
 	now func() time.Time
 
@@ -90,25 +93,35 @@ type PgxLeaderConfig struct {
 	SweepInterval   time.Duration
 	MaxPingAge      time.Duration
 	StepDownTimeout time.Duration
+	// OnIntegrityResult, if set, is called after every cluster integrity
+	// evaluation with the result (nil means healthy). It runs on the lead()
+	// goroutine and must not block or panic.
+	OnIntegrityResult func(err error)
+	// OnPingResult, if set, is called after every leader ping attempt with
+	// whether it succeeded and the time it ran. It runs on the lead()
+	// goroutine and must not block or panic.
+	OnPingResult func(success bool, t time.Time)
 }
 
 func NewPgxLeader(cfg PgxLeaderConfig) *PgxLeader {
 	return &PgxLeader{
-		store:           cfg.Store,
-		awb:             cfg.Broadcaster,
-		notify:          cfg.Notifier,
-		taskHandler:     cfg.TaskHandler,
-		chLeader:        cfg.LeaderChannel,
-		chFollower:      cfg.FollowerChannel,
-		chMessages:      cfg.MessagesChannel,
-		address:         cfg.Address,
-		stop:            cfg.StopChan,
-		ping:            cfg.PingInterval,
-		sweep:           cfg.SweepInterval,
-		maxPingAge:      cfg.MaxPingAge,
-		stepDownTimeout: cfg.StepDownTimeout,
-		now:             time.Now,
-		mutex:           sync.RWMutex{},
+		store:             cfg.Store,
+		awb:               cfg.Broadcaster,
+		notify:            cfg.Notifier,
+		taskHandler:       cfg.TaskHandler,
+		chLeader:          cfg.LeaderChannel,
+		chFollower:        cfg.FollowerChannel,
+		chMessages:        cfg.MessagesChannel,
+		address:           cfg.Address,
+		stop:              cfg.StopChan,
+		ping:              cfg.PingInterval,
+		sweep:             cfg.SweepInterval,
+		maxPingAge:        cfg.MaxPingAge,
+		stepDownTimeout:   cfg.StepDownTimeout,
+		onIntegrityResult: cfg.OnIntegrityResult,
+		onPingResult:      cfg.OnPingResult,
+		now:               time.Now,
+		mutex:             sync.RWMutex{},
 	}
 }
 
@@ -276,6 +289,9 @@ func (p *PgxLeader) clusterIntegrityErr() error {
 // (when that timeout is non-zero). Called from the lead() goroutine only.
 func (p *PgxLeader) evaluateHealth() bool {
 	err := p.clusterIntegrityErr()
+	if p.onIntegrityResult != nil {
+		p.onIntegrityResult(err)
+	}
 	if err == nil {
 		if !p.unhealthySince.IsZero() {
 			slog.Info("Cluster integrity recovered", "degradedFor", p.timeNow().Sub(p.unhealthySince))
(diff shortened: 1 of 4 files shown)


jonyoder (commented):
Self-review notes from a local pass (Claude Code). The main thing worth addressing before merge is the goroutine doc contract; the rest are minor.

pkg/rselection/impls/pgx/leader.go
```diff
@@ -90,25 +93,35 @@ type PgxLeaderConfig struct {
 	SweepInterval   time.Duration
 	MaxPingAge      time.Duration
 	StepDownTimeout time.Duration
+	// OnIntegrityResult, if set, is called after every cluster integrity
+	// evaluation with the result (nil means healthy). It runs on the lead()
+	// goroutine and must not block or panic.
+	OnIntegrityResult func(err error)
+	// OnPingResult, if set, is called after every leader ping attempt with
+	// whether it succeeded and the time it ran. It runs on the lead()
```
jonyoder: The "It runs on the lead() goroutine" claim is inaccurate for `OnPingResult`, and the inaccuracy understates concurrency. The periodic path dispatches `go p.pingNodes(ctx)` (leader.go:205), so this fires on a transient per-tick goroutine — only the startup ping at leader.go:174 runs on `lead()`. Net effect: `OnPingResult` (ping goroutine) and `OnIntegrityResult` (lead goroutine, sweep tick) run concurrently with each other. The current wording invites a consumer to assume serialized execution and skip synchronization on shared callback state. Suggest rewording both leader callbacks to promise what is actually guaranteed, e.g. "May be called concurrently with the other callbacks and across invocations; must be non-blocking, panic-free, and safe for concurrent use."

pkg/rselection/taskhandler.go
```diff
@@ -50,20 +50,26 @@ type Schedule interface {
...
+	verify    chan chan bool
+	mutex     sync.RWMutex
+	role      string
+	onTaskRun func(name string)
 }
 
 type GenericTaskHandlerConfig struct {
+	// OnTaskRun, if set, is called with the task name whenever a scheduled task
+	// is about to run (after the cluster verify gate passes). It runs on the
+	// task handler's goroutine and must not block or panic.
```
jonyoder: Same concurrency-contract issue here: `Handle()` launches `go h.runScheduled(...)` per scheduled task, so with N scheduled tasks there is no single "task handler's goroutine" and `OnTaskRun` can fire concurrently across tasks. Recommend rewording to state it may be called concurrently and must be safe for concurrent use.

pkg/rselection/impls/pgx/leader_recovery_test.go
```diff
@@ -141,3 +142,46 @@ func (s *RecoverySuite) TestStepDownDisabledByZeroTimeout(c *check.C) {
...
+
+func (s *RecoverySuite) TestPingNodesFiresPingCallback(c *check.C) {
+	var results []bool
+	leader := &PgxLeader{
+		address:    "leader",
+		chLeader:   "leader",
+		chFollower: "follower",
+		notify:     FakePgNotifier{},
+	}
+	leader.onPingResult = func(success bool, _ time.Time) { results = append(results, success) }
```
jonyoder: Minor test gap: the callback receives a `t time.Time` that nothing asserts on. Consider injecting `leader.now` (the test relies on the `timeNow()` fallback today) and asserting the timestamp is passed through, since that arg is part of the callback contract and is otherwise untested.

pkg/rselection/impls/pgx/leader.go
jonyoder: Agreed and fixed in 5de2848. You're right that only the startup ping runs on lead(); periodic pings dispatch `go p.pingNodes` (leader.go:205), so OnPingResult runs on transient per-tick goroutines and can overlap both itself and OnIntegrityResult. Reworded both leader callbacks to drop the lead()-goroutine claim and require non-blocking, panic-free, and safe-for-concurrent-use. (PPM's consumers only do goroutine-safe Prometheus ops, so no consumer change needed.)

pkg/rselection/taskhandler.go
jonyoder: Fixed in 5de2848 — reworded OnTaskRun to note that Handle() launches a goroutine per scheduled task, so it may be called concurrently across tasks and must be safe for concurrent use.

pkg/rselection/impls/pgx/leader_recovery_test.go
jonyoder: Fixed in 5de2848 — TestPingNodesFiresPingCallback now injects `leader.now` (a fixed clock) and asserts the timestamp is passed through to the callback on both the success and failure paths.

pkg/rselection/impls/pgx/leader.go
```diff
@@ -276,6 +292,9 @@ func (p *PgxLeader) clusterIntegrityErr() error {
 // (when that timeout is non-zero). Called from the lead() goroutine only.
 func (p *PgxLeader) evaluateHealth() bool {
 	err := p.clusterIntegrityErr()
+	if p.onIntegrityResult != nil {
+		p.onIntegrityResult(err)
+	}
```
CDRayn, line 297: Should this also check if `err` is not `nil` to avoid unnecessary reporting of errors that are actually `nil`?

CDRayn (commented):
Overall this looks good. I just had a single question about checking an error for `nil` before reporting the error.

pkg/rselection/impls/pgx/leader.go
jonyoder: Good question — this is intentional. `OnIntegrityResult` is a *result* hook, not a failure hook: it reports the outcome of every integrity check, where `nil` means healthy (as noted in its doc comment). That mirrors `OnPingResult`, which likewise fires on every ping with `success bool` (both true and false), so the two leader callbacks stay consistent.

Firing on `nil` also lets a consumer observe healthy checks, not just failures — e.g. recording a "last successful integrity check" timestamp or resetting state on recovery. Guarding with `err != nil` here would turn it into a failure-only hook and foreclose that.

The current consumer (PPM) treats the healthy case as a no-op (`if err != nil { ClusterIntegrityCheckFailed() }`), so a nil result never increments anything, and the per-sweep nil call is negligible. Happy to revisit if you'd prefer a narrower failure-only hook, but I'd lean toward keeping it symmetric with OnPingResult.

PR #290, opened 2026-07-18
Add GHA workflow to publish releases; stale file cleanup

glin:
Currently, platform-lib releases are done by manual tagging + creating a GH release. This adds a GHA workflow to automatically tag + release given a version.

Also cleans up some old release files that are no longer used.

### Release workflow

`.github/workflows/release.yml` publishes a GitHub Release with a generated changelog, two ways:

- Run it with a version string: `gh workflow run release.yml -f version=v4.3.1` (validates the version, refuses duplicates, requires `main`, then tags and releases from CI, no local git)
- Or push a semver tag: `git tag v4.3.1 && git push origin v4.3.1`

### Cleanup

The repo has been one module (`github.com/rstudio/platform-lib/v4`) since v4, so:

- Deleted the chart/tags scripts (`latest-tags.sh`, `generate-chart-data.sh`, `generate-chart.sh`), their justfile targets, and the `.chart` entries.
- Rewrote the README Release + Testing sections and the `pkg` READMEs; fixed test examples pointing at packages and import paths that no longer exist.

### Testing

Hard to test but after merge, I'll try publishing a release, then delete the release/tag (unless we have legit changes to include, possibly some dependency bumps)


jonyoder (approved):
Great to see this added!

CDRayn (approved):

PR #304, opened 2026-09-09
Fix Unsubscribe deadlock during shutdown

nodivbyzero:
### Description
The Unsubscribe methods in NotificationBroadcaster and DatabaseQueue blocked forever when called after the broadcaster's loop had exited.

During shutdown, if the broadcaster stops before a subscriber calls Unsubscribe, the send to the unsubscribe channel has no receiver and blocks forever, causing a goroutine leak.

Fix by using a select that also checks if the broadcaster has stopped (the stop channel is closed when the loop exits). When the broadcaster has already stopped, all channels were closed by stop(), so the drainer goroutine will exit naturally when it sees the closed channel.

Add unit tests that verify Unsubscribe returns immediately after the broadcaster has stopped.

Connected to: https://github.com/rstudio/package-manager/issues/19987

### DIFF
--- a/pkg/rsnotify/broadcaster/broadcaster.go
+++ b/pkg/rsnotify/broadcaster/broadcaster.go
@@ -189,7 +189,12 @@ func (b *NotificationBroadcaster) Unsubscribe(ch <-chan listener.Notification) {
 		}
 	}
 	go drainer()
-	b.unsubscribe <- ch
+	select {
+	case b.unsubscribe <- ch:
+	case <-b.stopSignal:
+		// Broadcaster has stopped; all channels were closed by stop().
+		// The drainer will exit when it sees the closed channel.
+	}
 }
 
 // internal stop function that closes the destination channels.
--- a/pkg/rsnotify/broadcaster/broadcaster.go
+++ b/pkg/rsnotify/broadcaster/broadcaster.go
@@ -183,6 +183,9 @@ func (b *NotificationBroadcaster) SubscribeOne(dataType uint8, matcher Matcher) 
 // Unsubscribe removes a channel from receiving broadcast events. That channel is
 // closed as a consequence of unsubscribing.
 func (b *NotificationBroadcaster) Unsubscribe(ch <-chan listener.Notification) {
+	if ch == nil {
+		return
+	}
 	drainer := func() {
 		// It's possible that the broadcaster is still trying to send
 		// us events while we're attempting to unsubscribe. Create a
--- a/pkg/rsnotify/broadcaster/broadcaster.go
+++ b/pkg/rsnotify/broadcaster/broadcaster.go
@@ -142,15 +142,20 @@ type Subscription struct {
 }
 
 // Subscribe returns a new output channel that will receive all broadcast events.
+// Returns nil if the broadcaster has already stopped.
 func (b *NotificationBroadcaster) Subscribe(dataType uint8) <-chan listener.Notification {
 	c := make(chan listener.Notification)
 
-	b.subscribe <- Subscription{
+	select {
+	case b.subscribe <- Subscription{
 		C: c,
 		T: dataType,
+	}:
+		return c
+	case <-b.stopSignal:
+		// Broadcaster has stopped; return nil to indicate no subscription.
+		return nil
 	}
-
-	return c
 }
 
 // SubscribeOne returns a new output channel that will receive one and only one broadcast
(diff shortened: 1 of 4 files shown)


jonyoder:
Nice fix — this closes both goroutine leaks reported in rstudio/package-manager#19987.

One thing worth a look before merging: `Subscribe`/`SubscribeOne` have the same unconditional-send shape that `Unsubscribe` had here, just unfixed.

- In `broadcaster.go`, `Subscribe` and `SubscribeOne` (~lines 145–171) do `b.subscribe <- Subscription{...}` with no select against `b.stopSignal`.
- In `queue.go`, `SubscribeOne` (~lines 200–210) does the same against `q.stopChan`.

If the broadcaster's loop has already exited when one of these is called, the send blocks forever — identical leak shape to the one just fixed, just on the other side of the channel pair.

This isn't hypothetical: `DatabaseQueue.Get` (queue.go:378) and the `PollAddress` poll loop (queue.go:288, 294) call `SubscribeOne` on every iteration for the life of an agent, so there's a real window during shutdown where this could fire. To be fair, the two leaks in #19987 both landed on `Unsubscribe`, not `Subscribe`, so I don't have evidence this is happening in practice today — but it's the same bug in the same files, and the fix is the identical one-line pattern already applied above. Worth extending it to `Subscribe`/`SubscribeOne` here, or tracking as a quick follow-up if you'd rather keep this PR scoped to the reported leaks.

nodivbyzero:
> Nice fix — this closes both goroutine leaks reported in [rstudio/package-manager#19987](https://github.com/rstudio/package-manager/issues/19987).
> 
> One thing worth a look before merging: `Subscribe`/`SubscribeOne` have the same unconditional-send shape that `Unsubscribe` had here, just unfixed.
> 
>     * In `broadcaster.go`, `Subscribe` and `SubscribeOne` (~lines 145–171) do `b.subscribe <- Subscription{...}` with no select against `b.stopSignal`.
> 
>     * In `queue.go`, `SubscribeOne` (~lines 200–210) does the same against `q.stopChan`.
> 
> 
> If the broadcaster's loop has already exited when one of these is called, the send blocks forever — identical leak shape to the one just fixed, just on the other side of the channel pair.
> 
> This isn't hypothetical: `DatabaseQueue.Get` (queue.go:378) and the `PollAddress` poll loop (queue.go:288, 294) call `SubscribeOne` on every iteration for the life of an agent, so there's a real window during shutdown where this could fire. To be fair, the two leaks in #19987 both landed on `Unsubscribe`, not `Subscribe`, so I don't have evidence this is happening in practice today — but it's the same bug in the same files, and the fix is the identical one-line pattern already applied above. Worth extending it to `Subscribe`/`SubscribeOne` here, or tracking as a quick follow-up if you'd rather keep this PR scoped to the reported leaks.

Nice catch.
`subscribe/SubscribeOne` is fixed in 787e91cf86b8c934ff30be3d56759d1e793f7015

jonyoder:
Thanks for the quick turnaround — the direction is right (mirrors the `Unsubscribe` pattern), but `787e91c` introduces a new leak of the same kind it's fixing.

`Subscribe`/`SubscribeOne` now return `nil` once the broadcaster has stopped, but none of the existing callers check for that before deferring `Unsubscribe`:

- `queue.go` `Get()`: `qAvail := q.SubscribeOne(...); defer q.Unsubscribe(qAvail)`
- `queue.go` `PollAddress`'s inner closure: same pattern for `completedMsgs`/`chunkMsgs`
- `tasks/monitor.go` `DatabaseQueueMonitorTask.Run`: `sub := b.Subscribe(...); defer b.Unsubscribe(sub)`

`Unsubscribe`'s `drainer` goroutine does `_, more := <-ch` unconditionally, with no nil check. Receiving from a nil channel never becomes ready, so `Unsubscribe(nil)` returns fine (the `<-stopChan` case fires), but the `drainer` goroutine it spawned is now parked forever on `<-nil` — a permanent leak.

Reproduced this locally: subscribe-after-stop returns `nil`, then calling `Unsubscribe` on that `nil` channel trips `leaktest`:

```
leaktest: leaked goroutine: goroutine 83 [chan receive (nil chan)]:
github.com/rstudio/platform-lib/v4/pkg/rsnotify/broadcaster.(*NotificationBroadcaster).Unsubscribe.func1()
	pkg/rsnotify/broadcaster/broadcaster.go:193
created by .../Unsubscribe
	pkg/rsnotify/broadcaster/broadcaster.go:201
```

Same thing reproduces in `database.(*DatabaseQueue).Unsubscribe` at queue.go:222/228.

Suggest fixing it once, at the source: have `Unsubscribe` itself guard `if ch == nil { return }` before starting the drainer/select, in both `broadcaster.go` and `queue.go`. That protects the existing callers without having to touch each call site, and covers any future caller that doesn't think to check the `Subscribe`/`SubscribeOne` return value either.

jonyoder:
Confirmed — `b85419a`'s nil guard fixes it. Re-ran my repro against this commit and it's clean, and the full package tests pass (including `-race`).

(Unrelated: `-race -count=3` surfaces a data race in `TestPollTickOk` via `fakeMetrics.QueueNotificationMiss`, but I checked and it reproduces on `main` too, so it predates this PR — not something to fix here.)

LGTM.

jonyoder (approved):
Looks good! Thanks for these fixes.

